Issue #3, opened 2018-10-08
Data questions


beamalsky:
An issue for making note of questions about the hearing data as they comes up.

- Do the Congress ID numbers (eg, the 2017-2018 Congress is referred to as 115) need to be retained? Is it important to be able to referred to sessions of Congress in two-year increments?
- Does the party of a committee chair always correspond to the majority party in the House and Senate?
- Do chairs and Congress majorities always change cleanly at the end of the year?

beamalsky:
@derekeder fyi for tomorrow, this is my running list of questions for Jamie/Lugar ^^

PR #5, opened 2018-10-17
Form setup

beamalsky:
Questions for Lugar logged in #3. Next steps: better validation, some tests would be very nice!

It occurred to me midway through this that maybe this should be in the admin interface. LMK if you feel that way too.


committeeoversight/local_settings.py
```diff
@@ -0,0 +1,22 @@
+# SECURITY WARNING: keep the secret key used in production secret!
+# Set this to an arbitrary string
+SECRET_KEY = '*mb&wj8q@n5jy!fki)qi*%@i=+gc7v1$a7*v*=bj&xktxi0g&b'
```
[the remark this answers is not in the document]
beamalsky: Ee thanks for catching!

committeeoversightapp/views.py
```diff
@@ -1,3 +1,160 @@
...
+                name = committee.name
+                event = event_form.save(commit=False)
+                organization = Organization.objects.get(id=committee.id)
+                entity_type = "organization"
+                new_committee = EventParticipant(name=name, event=event, organization=organization, entity_type=entity_type)
+                new_committee.save()
+
+            # find and create committee members as EventParticipants
+            committeemembers = committeemember_form.cleaned_data['name'].split(",")
+
```
reginafcompton: I am not persuaded that this is the best way to do this, i.e., by asking the user to input a comma-separated list. (Is there a reason you've implemented it that way? did the folks at Lugar suggest something along those lines?)

This seems like a good opportunity for a [formset](https://docs.djangoproject.com/en/2.1/topics/forms/formsets/), which would allow the user to create several committee members and witnesses (helping you avoid the awkwardness of cleaning up after calling `split`). Happy to discuss this face-to-face!

committeeoversightapp/views.py
```diff
@@ -1,3 +1,160 @@
...
+            else:
+                note="transcript"
+                new_document = EventDocument(note=note, event=event)
+                new_document.save()
+
+                archived_transcript_url = archive_url(transcript_url)
+                print(archived_transcript_url)
+
+                ext = get_ext(transcript_url)
+                if ext.lower() == '.pdf':
```
reginafcompton: You could make this a tad more DRY. One way that comes to mind: a mapping of the extensions, rather than the successive IF/ELSE statements. So, something like: 

```
extensions = {'.pdf': 'application/pdf', '.htm': 'text/html'} 
...
media_type=extensions[ext.lower()]
```

beamalsky:
Yes! Per the scope, "Lugar Center staff should be able to create accounts, log in, and save entries"

README.md
```diff
@@ -26,3 +26,37 @@ Perform the following steps from your terminal.
...
     pip install -r requirements.txt
     ```
+
+4. Copy the example local settings file to the correct location:
+
+  ```bash
+  cp committeeoversight/local_settings.example.py committeeoversight/local_settings.py
+  ```
+
+5. Make the database:
```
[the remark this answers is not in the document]
beamalsky: Makes sense! I'll put in a placeholder and make sure to updated once we've worked out with Forest how the scraper will fit in here.

committeeoversightapp/forms.py
```diff
@@ -0,0 +1,81 @@
...
+        widget=Textarea(attrs={'rows':4, 'cols':15}),
+        required=False
+    )
+
+# add a field for transcript url
+class TranscriptForm(Form):
+    url = CharField(
+        label='Transcript URL',
+        required=False
+    )
```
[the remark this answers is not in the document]
beamalsky: Copying over to #13

committeeoversightapp/views.py
beamalsky: I'd started with a formset but it seemed clunky if the data enterer is started with a single string. I asked about formatting for this and the Witnesses field in my email to Jamie this morning, will adjust accordingly!

committeeoversightapp/views.py
beamalsky: oo clever!

beamalsky:
@reginafcompton thanks! I've responded to your comments, and I think I'll handle bettering the validation in a separate PR.

committeeoversightapp/views.py
```diff
@@ -118,32 +118,23 @@ def post(self, request, **kwargs):
...
-                                            media_type="application/pdf"
-                                        )
-                    new_archived_document_link = EventDocumentLink(
-                                            url=archived_transcript_url,
-                                            document=new_document,
-                                            media_type="application/pdf"
-                                        )
-                elif ext.lower() == '.htm' or ext == '.html':
+
+                try:
```
reginafcompton: Since the CharField returns an empty string, you can just do:

```
media_type = extensions.get(ext.lower(), '')
```

Then, you can remove the Try-Except business.

committeeoversightapp/views.py
reginafcompton: Terrific. I think let's leave this for now, but can you open an issue about it? Then, we will hear from Jamie, we'll decide the best course of action.

committeeoversightapp/views.py
beamalsky: Much better, thanks!

PR #16, opened 2018-10-29
Form data updates

beamalsky:
Formset reference: https://docs.google.com/presentation/d/1NefIX7rv0GTgI9SBbxVkCEGOpekUw72LDjdDJRewO2U/edit#slide=id.g2d5227ca3b_0_29

https://github.com/elo80ka/django-dynamic-formset/blob/master/docs/usage.rst


beamalsky:
Refigured the forms based on discussion with Jamie. As it stands: 

1. Title*
2. Date*
3. Committee or subcommittee* (at least one, possible to have multiple)
4. Category* (1-8)
5. Transcript URL
7. Witness(es): fields for name*, witness organization, and witness statement URL (possible to have multiple, but each witness has no more than one of each)

There are a couple new models to handle the need for another category for the hearings and a way to link witnesses with organizations and documents. Waiting to hear back from Jamie on how I should add fields for opening statement transcripts!

beamalsky:
@reginafcompton I haven't done a huge amount of QA here yet—I'd like to get the edit functionality working first, which will make it much easier keep track of things.

committeeoversightapp/fixtures/hearingcategorytype.json
```diff
@@ -0,0 +1,58 @@
+[
```
reginafcompton: Can you tell me a bit about the purpose of the JSON file?

committeeoversightapp/forms.py
```diff
@@ -1,45 +1,36 @@
...
         }
         widgets = {
-            'jurisdiction': HiddenInput()
+            'jurisdiction': HiddenInput(),
+            'classification': HiddenInput()
             }
 
     start_date = DateField(
-        widget=DateInput(attrs={'placeholder': 'yyyy-mm-dd'})
+        widget=DateInput(attrs={'placeholder': 'yyyy-mm-dd'}),
```
reginafcompton: Not terribly important at the moment, but you might look into using [jquery's datepicker](https://jqueryui.com/datepicker/) for this field.

committeeoversightapp/models.py
```diff
@@ -1 +1,18 @@
...
+
+class HearingCategoryType(models.Model):
+    id = models.CharField(max_length=100, primary_key=True)
+    name = models.CharField(max_length=100, primary_key=False)
+
+    def __str__(self):
+        return u'({0}) {1}'.format(self.id, self.name)
+
+class HearingCategory(models.Model):
+    event = models.ForeignKey(Event, null=True, on_delete=models.CASCADE)
```
reginafcompton: ah, `on_delete` is now a required argument for Django 2.....

committeeoversightapp/static/css/custom.css
```diff
@@ -10,6 +10,44 @@ body {
...
+  float: right;
+  padding: 15px 0px;
+  font-size: 14px;
+  font-weight: 500;
+}
+
+.delete-row:hover {
+  color: #6f0017;
+}
+
```
reginafcompton: custom styles - nice! I always find that I can minimize CSS-refactor-headaches by putting all colors inside variables: https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables and [an example](https://github.com/datamade/nyc-council-councilmatic/blob/master/nyc/static/css/city_custom.css#L1)

committeeoversightapp/views.py
```diff
@@ -36,8 +38,41 @@ def archive_url(url):
...
     archived = requests.get(save_url)
-    archive_url = '{0}{1}'.format(wayback_host, archived.headers['Content-Location'])
-    return archive_url
+    try:
+        archive_url = '{0}{1}'.format(wayback_host, archived.headers['Content-Location'])
+        return archive_url
+    except KeyError:
+        return None
+
+def save_document(url, note, event):
```
reginafcompton: I'd put this function (and the others above it) in a `utils.py` file, since they are not technically views.

committeeoversightapp/fixtures/hearingcategorytype.json
beamalsky: Lugar has 8 categories of hearing, and they'd like to assign one to each hearing as it's entered. I talked to Forest and these don't fit well into the OCD event model, so I've made two new models to house the categories and their association with hearings: HearingCategory and HearingCategory type. The categories should be a pre-populated drop down menu in the form, and this fixture lets me load the data into the app.

committeeoversightapp/forms.py
beamalsky: On my list of questions for Lugar is whether or not we can depend on this always being an exact day!

committeeoversightapp/models.py
beamalsky: the only world I've ever known...

committeeoversightapp/static/css/custom.css
beamalsky: love it, added!

committeeoversightapp/forms.py
beamalsky: Opening an issue + closing

committeeoversightapp/migrations/0001_initial.py
```diff
@@ -0,0 +1,48 @@
...
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+        ('legislative', '0008_merge_20181031_1341'),
```
reginafcompton: This points to a merge migration that, I believe, only resides on your local machine. (You can take a look at `opencivicdata/legislative/migrations` in your virtual environment, and I suspect you will see it there.) 

In any case, the dependency for this migration should be: https://github.com/opencivicdata/python-opencivicdata/blob/master/opencivicdata/legislative/migrations/0007_auto_20181029_1527.py

Can you look into this (erp, again)?

reginafcompton: I think this migration conflict concerns the hearings scraper, which points to a particular branch on python-opencivicdata: https://github.com/datamade/hearings/blob/master/requirements.txt#L4 

Let's touch base with @fgregg about whether he plans to merge that branch with master. 

In any case, I think we want the `committee-oversight` and `hearings` repos to point to the same version of python-opencivicdata (i.e., to avoid `django.db.migrations.exceptions.NodeNotFoundError`). 'Til Tuesday!

beamalsky: @fgregg: as of last week there's a new version of OCD that has some migration changes: https://github.com/opencivicdata/python-opencivicdata/commit/3fcbae298b95311b1d9b02faf9a86704f1632234

I updated this `committee-hearings` app and it's causing some merge conflicts. We should make sure this app and the hearings scraper repo use the same version. What do you think?

fgregg: you need to point the app to the same python-opencividata as hearings.

PR #20, opened 2018-11-06
Add list view

beamalsky:
A quick list view!


committeeoversightapp/templates/list.html
```diff
@@ -0,0 +1,36 @@
...
+
+{% block content %}
+<div class="hearing-form card">
+
+ <p><a href="/" class="btn btn-success">Create new hearing</a></p>
+
+ <table class="table table-striped" id="hearing-table">
+  <thead>
+    <tr>
+      <th scope="col" style="width: 60%">Hearing Title</th>
```
reginafcompton: This is normally how I set table column width, too! But I was curious if Bootstrap (4) had a way to accomplish the same thing. Yes, it does: https://stackoverflow.com/questions/37924104/table-column-sizing

beamalsky: elegant!

beamalsky:
@reginafcompton Ok this is updated and merged unconflicted!

PR #24, opened 2018-11-07
Styling

beamalsky:
Some styling changes, subject to change after a design review meeting with Derek tomorrow

### DIFF
--- a/committeeoversightapp/templates/create.html
+++ b/committeeoversightapp/templates/create.html
@@ -50,7 +50,10 @@
       {% endfor %}
     </div>
 
-    <button type="submit" class="btn-lg btn-success float-right" id="save-hearing-button">Save hearing</button>
+    <button type="submit" class="btn-lg btn-success float-right" id="save-hearing-button">
+      <i class="far fa-save"></i>&nbsp; 
+      Save hearing
+    </button>
 
   </form>
 </div>
--- a/committeeoversightapp/static/css/custom.css
+++ b/committeeoversightapp/static/css/custom.css
@@ -94,13 +94,17 @@ body {
     border-color: var(--dark-blue-color);
 }
 
+.page-link {
+  color: var(--dark-blue-color);
+}
+
 .navbar {
     background: var(--dark-blue-color);
 }
 
 .navbar div li {
   color: white;
-  padding: 10px 20px;
+  padding: 10px 15px;
   font-size: .95rem;
 }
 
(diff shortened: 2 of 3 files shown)


committeeoversightapp/templates/list.html
```diff
@@ -5,10 +5,10 @@
 
  <p><a href="/" class="btn btn-success">Create new hearing</a></p>
 
- <table class="table table-striped" id="hearing-table">
+ <table class="table table-striped table-bordered" id="hearing-table">
   <thead>
     <tr>
-      <th scope="col" class="w-75">Hearing Title</th>
+      <th scope="col" id="hearing-title-col">Hearing Title</th>
```
reginafcompton: nice - this works, too!

committeeoversightapp/templates/create.html
```diff
@@ -50,12 +50,9 @@ <h4>Witnesses</h4>
       {% endfor %}
     </div>
 
-    <br />
-
-    <button type="submit" class="btn-lg btn-success float-right">Save hearing</button>
+    <button type="submit" class="btn-lg btn-success float-right" id="save-hearing-button">Save hearing</button>
```
reginafcompton: If you are in front-end mode, you might consider adding a couple Font Awesome icons, particularly for the buttons.

committeeoversightapp/templates/list.html
beamalsky: I wanted it more precise than the w-*s allowed! A useful trick for the future, though.

committeeoversightapp/templates/create.html
beamalsky: Great pointed, added some

committeeoversightapp/templates/list.html
```diff
@@ -29,6 +29,7 @@
 $(document).ready( function () {
     $('#hearing-table').DataTable({
       "dom": "Bfrtip",
+      "order": [[ 1, "desc" ]]
```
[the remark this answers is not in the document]
beamalsky: The line in `views.py` wasn't working, so I switched it over to here. I didn't do much troubleshooting, though. Is one better than the other?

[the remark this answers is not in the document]
reginafcompton: Ah, I see. Unless you disable ordering (`"order": []`), then dataTables initializes the table with a default order (`[[0, 'asc']]`), which would override the ListView ordering. https://datatables.net/reference/option/order#Default 

I think it makes more sense to order in the dataTables, as you're doing!

PR #26, opened 2018-11-09
Add datepicker type

### LINKED ISSUE #17: Add date picker
Check that hearings will always have exact dates --> if so, add [jquery's datepicker](https://jqueryui.com/datepicker/)

beamalsky:
Well this was more straightforward than I thought! Closes #17


beamalsky:
I like the Django datepicker because it guides you through typing mm/dd/yyyy and only allows valid dates!

reginafcompton:
Great - then, let's use that!

PR #28, opened 2018-11-09
Make this friendlier on mobile

beamalsky:
A couple styling changes to make this smooth on small screens


committeeoversightapp/templates/base.html
```diff
@@ -30,9 +30,9 @@
 
   <body>
 
-    <nav class="navbar fixed-top footer navbar-expand-sm">
+    <nav class="navbar navbar-dark fixed-top footer navbar-expand-md">
```
reginafcompton: well, that makes a difference!

PR #32, opened 2018-11-21
Add Boolean 'retired' value to WitnessDetails model and form

beamalsky:
Defaults to False


committeeoversightapp/views.py
```diff
@@ -78,7 +78,8 @@ def post(self, request, **kwargs):
                                         name=name,
                                         event=event,
                                         entity_type=entity_type,
-                                        note=note)
+                                        note=note
+                    )
```
reginafcompton: oh, details

requirements.txt
```diff
@@ -4,7 +4,7 @@ sqlalchemy==1.2.12
 psycopg2==2.7.5
 csvkit==1.0.3
 
-https://github.com/opencivicdata/python-opencivicdata/zipball/longer_event_names
+https://github.com/opencivicdata/python-opencivicdata/zipball/master
```
reginafcompton: I hesitate to pin this to master, since (as we saw) this repo receives a fair amount of attention from its maintainers. 

You can pin the requirements to a particular commit, or wait for release 2.3 (the commit history indicates that 2.3 is coming very soon: https://github.com/opencivicdata/python-opencivicdata/commits/master).

reginafcompton: For this PR, can you pin it to the most recent commit, and then open an issue with a reminder to change the requirement to 2.3, when it comes out? (Just "watch" the ocd repo.)

committeeoversightapp/views.py
beamalsky: 😈

requirements.txt
beamalsky: Done! The OCD line in `requirements.txt` is switched to a download URL I copied from Github. Let me know if there's a more elegant way to do this.

PR #37, opened 2018-11-27
Add source row to detail view

beamalsky:
We're adding source links to the hearing detail view to help eventually improve scraper handling of redundant hearings!

### DIFF
--- a/committeeoversightapp/templates/detail.html
+++ b/committeeoversightapp/templates/detail.html
@@ -38,7 +38,7 @@
    <tr>
      <td><b>Ranking Member Opening Statement URL</b></td>
      {% if opening_statement_rm_archived %}
-       <td><a href={{opening_statement_rm_archived}}>{{opening_statement_rm}}</a></td>
+       <td><a href={{opening_statement_rm_archived}}>{{opening_statement_rm}}/a></td>
      {% else %}
        <td><a href={{opening_statement_rm}}>{{opening_statement_rm}}</a></td>
      {% endif %}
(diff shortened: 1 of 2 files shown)


committeeoversightapp/templates/detail.html
```diff
@@ -47,5 +47,13 @@
      <td><b>Witnesses</b></td>
      <td>{{witnesses}}</td>
    </tr>
+   <tr>
+     <td><b>Sources</b></td>
+     <td>
+       {% for source in hearing.sources.all %}
+        <a href={{source}}>{{source}}<br /></a>
```
reginafcompton: Would it be useful to add "target=_blank"?

beamalsky: Yes! I've added this to the transcript links as well.

beamalsky:
@reginafcompton snuck one more think in here: adding hearing dates to the list view, as Jamie requested

committeeoversightapp/templates/list.html
```diff
@@ -1,29 +1,31 @@
 {% extends 'base.html' %}
 
 {% block content %}
-<div class="col col-lg-11">
+<div class="col">
```
reginafcompton: Is this a bootstrap class? I thought cols needed some kind of designation, e.g., col-md-12? Asking for my own learning!

beamalsky: `.col` will take up all the space it can! https://getbootstrap.com/docs/4.1/layout/grid/#all-breakpoints

reginafcompton: neat. thanks, bootstrap 4.

PR #39, opened 2018-12-04
Limit listed hearings to 500

beamalsky:
In preparation for a full scrape on the staging site


committeeoversightapp/views.py
```diff
@@ -103,6 +103,7 @@ class EventList(LoginRequiredMixin, ListView):
     model = Event
     template_name = 'list.html'
     context_object_name = 'hearings'
+    queryset = Event.objects.all().order_by('-created_at')[:500]
```
fgregg: this should probably be done in in template https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#slice

PR #41, opened 2018-12-10
Data import

beamalsky:
Start of a reproducible data pipeline for the spreadsheet import. Still need to figure out how to import committees! Some questions for Jamie:

1. In the Hearings spreadsheet, what does the text in "Hearing #" refer to? Does this need to be saved?
2. How should we handle full committee meetings consistently across data sources?
3. Should the hyperlinks in the Senate file be saved as sources? Looks like links might be missing from the House sheet.


beamalsky:
@reginafcompton what part of this data would you recommend storing in the repo?

beamalsky:
NB: I moved the slicing back to the view code because I realized the list view was grabbing the first 500 hearings before ordering them, which defeated my intention of displaying the most recently entered hearings.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -101,26 +89,26 @@ def add_senate_hearings(self):
...
+                        # save committee data (TK)
+                        # for committee in committees:
+
+                        # save event source
+                        source = EventSource(event=event, note="spreadsheet", url=source)
+                        source.save()
+
+                        # save category data
+                        try:
+                            category = HearingCategory(event=event, category_id=category)
```
reginafcompton: Would it be worthwhile to add an `updated_at` and/or `created_at` field to the Hearing Category model? 

Also, do we need to indicate the source of the HearingCategory, or do you think it's okay to reply on the source of its related Event?

committeeoversightapp/management/commands/import_data.py
```diff
@@ -101,26 +89,26 @@ def add_senate_hearings(self):
...
+                            category.save()
+                        except IntegrityError:
+                            bad_rows.append("Unrecognized category on Senate row " + str(row_count) + ": " + event.name)
```
reginafcompton: Can you tell me a little about this `bad_rows.txt` file? Is it something that we plan to share with Jamie?

committeeoversightapp/management/commands/import_data.py
```diff
@@ -18,19 +17,6 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
```
reginafcompton: Is there any reason why we'd want to import the `house` and `senate` hearings separately (e.g., importing just the house hearings)? If so, it would be worth adding optional arguments: https://docs.djangoproject.com/en/2.1/howto/custom-management-commands/#accepting-optional-arguments

with that said.....I do not *think* there's a reason for this, but thought I'd run it by you, just in case.

beamalsky:
Thanks @reginafcompton, I'll look at your feedback shortly!

The last commit creates a Committee model and imports what it can with names grabbed from Organization and OrganizationNames and with the correct parent_id. Unfortunately, the committee names from Lugar don't match up very well with the names from the scraper. I'm thinking we should ask them to adjust them or see where they got them.

A generated report on the specific committees that it can't import: 
[bad_rows.txt](https://github.com/datamade/committee-oversight/files/2669533/bad_rows.txt)

committeeoversightapp/management/commands/import_data.py
beamalsky: Right now all HearingCategory sources are the same as the source of the Event itself. This may change once Forest starts on the ML piece of this, so I think this is a question for him.

committeeoversightapp/management/commands/import_data.py
beamalsky: Not at the moment! There were only 3 hours allocated for this import so I think it makes sense to not implement this until we need to.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -62,26 +48,28 @@ def add_house_hearings(self):
...
+                    event, created = Event.objects.get_or_create(jurisdiction_id = jurisdiction_id,
                                     name = name,
                                     start_date = start_date,
                                     classification = classification)
-                    event.save()
 
-                    # save committee data (TK)
-                    # for committee in committees:
+                    if created:
+                        event.save()
```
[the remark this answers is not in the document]
beamalsky: Good heads up thanks!

README.md
```diff
@@ -33,7 +33,11 @@ Perform the following steps from your terminal.
     cp committeeoversight/local_settings.example.py committeeoversight/local_settings.py
     ```
 
-5. Start a `hearings` database with a partial scrape from the [hearings repo](https://github.com/datamade/hearings) or a data dump from a pal.
+5. Start a `hearings` database with a scrape from the [hearings repo](https://github.com/datamade/hearings) or a data dump from a pal. If you're following the latter path and have a `hearings.dump` file in the root of this project, restore it by running:
+
+  ```bash
+  pg_restore -C -j4 --no-owner hearings.dump | psql
+  ```
```
reginafcompton: You might briefly note that `pg_restore` works only for db dumps in the archive format (as opposed to a text format). Maybe link to the documentation? https://www.postgresql.org/docs/10/app-pgrestore.html

bad_rows.txt
```diff
@@ -0,0 +1,782 @@
+House Committees
```
reginafcompton: Do we want this file under version control?

committeeoversightapp/management/commands/import_data.py
```diff
@@ -0,0 +1,210 @@
...
+                committee_key = row['Committees and Subcommittees']
+                committee_name = row['b']
+                classification = "committee"
+
+                if committee_key:
+                    try:
+                        if committee_name == "Full Committee" or committee_name == "Full Commission":
+                            pass
+                        else:
+                            if len(committee_key) == 3:
```
reginafcompton: Can you add a comment about why this is a reliable strategy? (You are just looking for committees with key, like, "102" or "102," right?)

README.md
```diff
@@ -33,7 +33,11 @@ Perform the following steps from your terminal.
     cp committeeoversight/local_settings.example.py committeeoversight/local_settings.py
     ```
 
-5. Start a `hearings` database with a partial scrape from the [hearings repo](https://github.com/datamade/hearings) or a data dump from a pal.
+5. Start a `hearings` database with a scrape from the [hearings repo](https://github.com/datamade/hearings) or a data dump from a pal. If you're following the latter path and have a `hearings.dump` file in the root of this project, restore it by running:
+
+    ```bash
+    pg_restore -C -j4 --no-owner hearings.dump | psql
```
reginafcompton: `pg_restore` expects the db dump to be in archive format (rather than text format). You might want to briefly explain this? Perhaps, link to the documentation? https://www.postgresql.org/docs/10/app-pgrestore.html

committeeoversightapp/management/commands/import_data.py
```diff
@@ -0,0 +1,210 @@
...
+                    elif committee2 and subcommittee2 == (committee2 + str(0)):
+                        new_committee(committee2, event, row_count)
+                    elif committee2:
+                        new_committee(subcommittee2, event, row_count)
+
+                    if created:
+
+                        # save event source
+                        source = EventSource(event=event, note="spreadsheet", url=source)
+                        source.save()
```
reginafcompton: You can simply call `create`, and then you do not need to build the object and save it in distinct steps. You can also start to remove some of this obvious in-line comments., e.g. "save event source" - the code speaks for itself.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -0,0 +1,210 @@
...
+from django.db.models import Q
+
+from opencivicdata.core.models import Organization, OrganizationName
+from opencivicdata.legislative.models import Event, EventSource, EventParticipant
+from committeeoversightapp.models import HearingCategory, Committee
+
+jurisdiction_id = 'ocd-jurisdiction/country:us/legislature'
+bad_rows = []
+
+def new_committee(committee_key, event, row_count):
```
reginafcompton: Is there a reason to have this function outside of the Command class?

bad_rows.txt
beamalsky: No, adding to .gitignore

committeeoversightapp/management/commands/import_data.py
beamalsky: Added! Full committees are 3 digits and subcommittees are 4 or 5

committeeoversightapp/management/commands/import_data.py
beamalsky: Hmmm I was getting a `NameError` when I had it inside. Do you know why?

beamalsky:
Last week I added something to check for Lugar committees that matched multiple possible scraper Organizations. Those aren't saved to the database—leaving for now until we work out the scraper committees with Jamie.

committeeoversightapp/management/commands/import_data.py
reginafcompton: Are you calling it with the `self` keyword?

```python
# declare function
def get_committee(self, committee_name, parent):
......

# call function
self.get_committee(committee_name, parent)
```

I was able to do this locally. Let me know if you need to talk through it.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -0,0 +1,213 @@
...
+from django.core.exceptions import ObjectDoesNotExist
+from django.conf import settings
+from django.db import IntegrityError
+from django.db.models import Q
+
+from opencivicdata.core.models import Organization, OrganizationName
+from opencivicdata.legislative.models import Event, EventSource, EventParticipant
+from committeeoversightapp.models import HearingCategory, Committee
+
+jurisdiction_id = 'ocd-jurisdiction/country:us/legislature'
```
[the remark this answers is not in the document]
beamalsky: zivxhe1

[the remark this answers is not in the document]
beamalsky: Thank you! Would you mind double checking my new commit?

committeeoversightapp/management/commands/import_data.py
```diff
@@ -0,0 +1,213 @@
...
+                committee_name = row['b']
+                classification = "committee"
+
+                if committee_key:
+                    try:
+                        # ignore subcommittees that are actually a full committee meeting
+                        if committee_name == "Full Committee" or committee_name == "Full Commission":
+                            pass
+                        else:
+                            # full committees are three digits
```
reginafcompton: I'd like to see a slightly more verbose comment here, explaining where those "three digits" come from and perhaps citing an example. Otherwise, you're just summarizing what the code already says.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -0,0 +1,213 @@
...
+def get_committee(committee_name, parent):
+    classification = "committee"
+    organization = Organization.objects.filter(Q(name__icontains=committee_name, parent=parent, classification=classification) |
+                Q(other_names__name__icontains=committee_name, parent=parent, classification=classification))
+    if organization.count() == 1:
+        return organization[0]
+    elif organization.count() == 0:
+        return IndexError
+    else:
+        return ValueError
```
reginafcompton: It would be nice to find a way to refactor this, so that you do not need to catch these errors, but instead (1) create the committee and (2) log to `bad_rows`. Right now, it's little abstruse what's going on. Not necessary, of course (since what you have works, and we're running short on hours), but perhaps a meaningful exercise.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -139,26 +138,26 @@ def add_house_hearings(self):
...
-                    event, created = Event.objects.get_or_create(jurisdiction_id = jurisdiction_id,
+                    event, created = Event.objects.get_or_create(jurisdiction_id = self.jurisdiction_id,
                                     name = name,
                                     start_date = start_date,
                                     classification = classification)
 
                     # save committee1 data
                     if committee1 and not subcommittee1:
-                        new_committee(committee1, event, row_count)
+                        new_committee(self, committee1, event, row_count)
```
reginafcompton: Be sure to call this like `self.new_committee(committee1, event, row_count)`.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -12,9 +12,6 @@
 from opencivicdata.legislative.models import Event, EventSource, EventParticipant
 from committeeoversightapp.models import HearingCategory, Committee
 
-jurisdiction_id = 'ocd-jurisdiction/country:us/legislature'
-bad_rows = []
-
 def get_committee(committee_name, parent):
```
reginafcompton: I'd like to see this as a class function, too - as you did with `new_committee`.

beamalsky:
Merging for now, will open a different PR for additional pre-import checks to avoid having redundant hearings between the scraper and manual imports

PR #46, opened 2019-01-28
Sophisticate the hearing matching

beamalsky:
Out of 1437 hearings in `hearings-partial.dump`, how many have a match in Lugar's manually entered spreadsheet?

Matching on name and date (#47): 164
Matching with above method (this PR): 1920


committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +110,165 @@ def add_house_hearings(self):
...
+                try:
+                    event = Event.objects.get(
+                                            name=name,
+                                            start_date=start_date,
+                                            jurisdiction_id=self.jurisdiction_id,
+                                            classification=classification
+                                            )
+                except ObjectDoesNotExist:
+                    # try to match by serial number
+                    if name and hearing_number_raw:
```
fgregg: I think we want to try to match by hearing_number first and then fall back to name and date.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +110,165 @@ def add_house_hearings(self):
...
+                        try:
+                            hearing_number = re.search(r'\d{2,}-\d{1,}', hearing_number_raw).group(0)
+                            matched_events = Event.objects.filter(extras__hearing_number__endswith=hearing_number)
 
-                    # save committee data
-                    for committee in committees:
-                        if committee:
-                            self.new_committee(committee, event, row_count)
+                            for event in matched_events:
```
fgregg: this code seems to contemplate that we have a one-to-many match.  please add a comment about why that is okay.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +109,165 @@ def add_house_hearings(self):
...
-                        try:
-                            hearing_category = HearingCategory.objects.create(event=event, category_id=category)
-                        except IntegrityError:
-                            self.bad_rows.append("Row " + str(row_count) + ": Unrecognized category " + category + " on " + event.name)
+                # get committees into a edited list format
+                # committee codes with a zero appended indicate "full committee"
+                # and only the full committee will be recorded as an event participant
+                # hearings with a subcommittee listed will only have the subcommittee saved,
+                # as the full committee is attached as a parent
+                committees_filtered = []
```
fgregg: this should be factored out into a method. maybe something like `_participating_committees` and it can take the row dict.

fgregg: (where *this* is lines 117-127)

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +109,165 @@ def add_house_hearings(self):
...
+        try:
+            hearing_category, _ = HearingCategory.objects.get_or_create(event=event, category_id=category)
+
+        except IntegrityError:
+            self.bad_rows.append("Row " + str(self.row_count) + ": Unrecognized category " + category + " on " + event.name)
+
+    def new_source(self, event, note, url):
+        source, _ = EventSource.objects.get_or_create(event=event, note=note, url=url)
+
+    def smart_match(self, matched, name, committees_filtered, start_date, category, source, classification):
```
fgregg: committees_filtered should be renamed to participating_committees

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +109,165 @@ def add_house_hearings(self):
...
+        # try to match by date and committees
+        print("Attempting to match " + name)
+        try:
+            committee_qs = Committee.objects.filter(lugar_id__in=committees_filtered)
+            committee_set = set(committee.organization.name for committee in committee_qs)
+            matched_events = Event.objects.filter(participants__name__in=committee_set, start_date=start_date).distinct()
+        except ValueError:
+            matched_events = []
+            self.bad_rows.append("Row " + str(self.row_count) + ": Bad committee value in " + str(committees_filtered))
```
fgregg: this is about where this method should end. the rest of the code in this current method is not trying to match, but updating the results of a match.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +109,165 @@ def add_house_hearings(self):
...
 
         except (ObjectDoesNotExist, AttributeError, ValueError) as e:
-            self.bad_rows.append("Row " + str(row_count) + ": Bad committee key " + committee_key)
+            self.bad_rows.append("Row " + str(self.row_count) + ": Bad committee key " + committee_key)
+
+    def new_category(self, event, category):
+        try:
+            hearing_category, _ = HearingCategory.objects.get_or_create(event=event, category_id=category)
+
+        except IntegrityError:
```
fgregg: How could there be an integrity error here?

committeeoversightapp/management/commands/import_data.py
fgregg: `smart_match` doesn't really help me understand what this method is doing, maybe `match on participating committees`?

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +109,165 @@ def add_house_hearings(self):
...
-        organization = Organization.objects.filter(Q(name__icontains=committee_name, parent=parent, classification=classification) |
-                    Q(other_names__name__icontains=committee_name, parent=parent, classification=classification))
-        if organization.count() == 1:
-            return organization[0]
-        elif organization.count() == 0:
-            return IndexError
+        organization_exact = Organization.objects.filter(name=committee_name, parent=parent, classification=classification)
+        organization_contains = Organization.objects.filter(name__icontains=committee_name, parent=parent, classification=classification)
+
+        if organization_exact.count() == 1:
```
fgregg: this should be moved in front of organization_contains.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +109,165 @@ def add_house_hearings(self):
...
+            return organization_exact[0]
         else:
```
fgregg: then you wouldn't need this else.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +109,165 @@ def add_house_hearings(self):
...
+                except IndexError:
+                    new_committee, _ = Committee.objects.get_or_create(lugar_id=committee_key, lugar_name=committee_name)
+                    self.bad_rows.append("Not recognized: " + committee_key + ", " + committee_name)
+            else:
+                self.bad_rows.append("Multiple possible committees for " + committee_key + ": " + committee_name)
 
-    def new_committee(self, committee_key, event, row_count):
+    def new_event_participant(self, committee_key, event):
         try:
             name = Committee.objects.get(lugar_id=committee_key).organization.name
```
fgregg: you don't need this line.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +109,165 @@ def add_house_hearings(self):
...
             organization = Committee.objects.get(lugar_id=committee_key).organization
             entity_type = "organization"
```
fgregg: pretty sure you don't need this line either.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +109,165 @@ def add_house_hearings(self):
...
             committee, created = EventParticipant.objects.get_or_create(name=name, event=event, organization=organization, entity_type=entity_type)
```
fgregg: pretty sure tthis line can just be 

EventParticipant.objects.get_or_create(event=event, organization=organization)

committeeoversightapp/management/commands/import_data.py
```diff
@@ -59,24 +63,18 @@ def add_house_committees(self):
...
-                            else:
-                                organization = self.get_committee(committee_name, parent)
-                            new_committee, _ = Committee.objects.get_or_create(lugar_id=committee_key, lugar_name=committee_name, organization=organization)
-                    except IndexError:
-                        new_committee, _ = Committee.objects.get_or_create(lugar_id=committee_key, lugar_name=committee_name)
-                        self.bad_rows.append("Not recognized: " + committee_key + ", " + committee_name)
-                    except ValueError:
-                        self.bad_rows.append("Multiple possible committees for " + committee_key + ": " + committee_name)
+                            organization = self.get_committee(committee_name, parent, committee_key)
+                        new_committee, _ = Committee.objects.get_or_create(lugar_id=committee_key, lugar_name=committee_name, organization=organization)
```
fgregg: it's confusing to create a new committee by using get_or_create?

committeeoversightapp/management/commands/import_data.py
```diff
@@ -115,98 +109,165 @@ def add_house_hearings(self):
...
+                                            name=name,
+                                            start_date=start_date,
+                                            jurisdiction_id=self.jurisdiction_id,
+                                            classification=classification
+                                            )
+                except ObjectDoesNotExist:
+                    # try to match by serial number
+                    if name and hearing_number_raw:
+                        try:
+                            hearing_number = re.search(r'\d{2,}-\d{1,}', hearing_number_raw).group(0)
```
fgregg: I'd factor out this chunk into a method called something like `match_by_hearing_number`

committeeoversightapp/management/commands/import_data.py
fgregg: I also think that the indentation level is wrong here. Right now you are only updating the data if hearing number match.

committeeoversightapp/management/commands/import_data.py
beamalsky: This is checking that the hearing wasn't already imported from the spreadsheet, so that this script can be rerun. Do you think hearing_number should still be checked first?

fgregg: if you want to check to see if an item has already been linked, then maybe also check for source?

committeeoversightapp/management/commands/import_data.py
beamalsky: Adding a comment! I intended to only update the hearing at this point if there's a matching hearing number in the database—if there's not, it will try the `smart_match` function. I'm fixing the logic here to reflect that by adding `not matched` to the next if statement. Would love to learn a cleaner way to do this!

committeeoversightapp/management/commands/import_data.py
beamalsky: Factored!

committeeoversightapp/management/commands/import_data.py
beamalsky: Elsewhere these are saved with name and entity type attached (there are other event participants that aren't organizations, like witnesses). Why do you suggest not including that here?

committeeoversightapp/management/commands/import_data.py
beamalsky: Good call

committeeoversightapp/management/commands/import_data.py
beamalsky: Ah, I see I'm doing redundant work to get the name. Fixed that, but I think I need to save it with both name and entity type information. Could you say more?

committeeoversightapp/management/commands/import_data.py
beamalsky: Thanks!

committeeoversightapp/management/commands/import_data.py
beamalsky: Some of the categories in Lugar's spreadsheet are not one of the set categories Jamie sent us  (see [fixture](https://github.com/datamade/committee-oversight/blob/master/committeeoversightapp/fixtures/hearingcategorytype.json)). 0 and 14 both show up a couple times.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -145,18 +151,30 @@ def add_senate_hearings(self):
...
+                    else:
+                        events = self.match_by_hearing_number(hearing_number_raw, name, category, source)
+                        if events is not None:
+                            # hearing numbers can be non-unique if a hearing has multiple sessions
+                            # these are recorded in the Lugar data as separate events
+                            for event in events:
+                                self.update_hearing(event, category, source)
+                                break
+
+                        event = self.match_by_date_and_participants(name, participating_committees, start_date, category, source, classification)
```
fgregg: should we really try this match path regardless of the return value of match_by_hearing_number?

beamalsky: true, we should stop once a hearing number is found! I'll update.

beamalsky: _Updated_

PR #47, opened 2019-01-30
Example of matching on title

beamalsky:
Out of 1437 hearings in `hearings-partial.dump`, how many have a match in Lugar's manually entered spreadsheet?

Matching on name and date (proof of work in this PR): 164

I haven't looked into this, but I suspect this number would be higher if we made it case insensitive and striped hearing titles of spaces and quotes.


beamalsky:
This exploratory branch has served its purpose and can be closed.

PR #53, opened 2019-03-06
Remove sorting on edit/delete buttons

beamalsky:
plus add nginx directive as described https://github.com/datamade/devops/issues/69!


committeeoversightapp/views.py
```diff
@@ -77,12 +77,6 @@ class EventListJson(BaseDatatableView):
     # define the columns that will be returned
     columns = ['updated_at', 'name', 'start_date', 'id', 'id']
 
-    # define column names that will be used in sorting
```
fgregg: why is this removed?

beamalsky:
I added `proxy_set_header X-Forwarded-Proto $scheme;` to the location block in the nginx file, per @reginafcompton's instructions

```
location / {
    ...
    proxy_set_header X-Forwarded-Proto $scheme;
    ...
  }
```

committeeoversightapp/views.py
beamalsky: Defining which columns are sortable is something that theoretically can be done in either the `django-datatables-view` package or in DataTables itself through the [orderable](https://datatables.net/reference/option/columns.orderable) parameter. I had tried doing it with the package with this line, taken straight from the documentation, but it didn't work! I switched to the DataTables method and it did.

Issue #45, opened 2019-01-23
Matching approach


fgregg:
1. For senate hearings, try to match on serial number
2. For house meetings, try to match on name of committee / subcommittee and date time.

beamalsky:
Four-hour authorization description from Forest: 

> My understanding is that these hours are for implementing a matching approach in in the script that imports Lugar's data into the database that we expect will be significantly better than matching on the title of the hearing. This is with the understanding that while the matching will be better, it will not be perfect.

beamalsky:
Caveat: Not all senate hearings in the scraped data or manually entered data have serial numbers. If that's missing on either end, I'm thinking I should then try to match on name of committee / subcommittee and date time. Does that make sense to you @fgregg?

fgregg:
Yep!

beamalsky:
@fgregg can this be closed?

fgregg:
yes

PR #54, opened 2019-03-11
fix import data issues

fgregg:
address #45


committeeoversightapp/management/commands/import_data.py
```diff
@@ -99,82 +102,162 @@ def add_senate_committees(self):
...
+            reader = enumerate(reader, 2)
+            # ignore rows with a missing name
+            reader = ((i, row) for i, row in reader if row['Hearing/Report'])
 
-            for row in reader:
+            for i, (self.row_index, row) in enumerate(reader, 1):
                 name = row['Hearing/Report']
                 start_date = row['Date'].split('T', 1)[0]
-                source = row['source']
+                source = csvfile.name
```
beamalsky: Why did you switch this to `csvfile.name`? This is saved in the url field of the source model, which I think isn't what we want on the front end: the site is expecting those to all be links when they exist. Maybe we could have different values for `note`, instead of making them all `spreadsheet` in this import.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -99,82 +102,162 @@ def add_senate_committees(self):
...
+                                                start_date,
+                                                participating_committees,
+                                                classification,
+                                                category,
+                                                source,
+                                                source_hash)
+
+
+            lugar_in_db = len(Event.objects.filter(sources__url=csvfile.name))
+            assert abs(lugar_in_db - i) < 80
```
beamalsky: Where is 80 coming from? Could you leave a comment here?

committeeoversightapp/management/commands/import_data.py
```diff
@@ -196,12 +279,18 @@ def get_committee(self, committee_name, parent, committee_key):
         else:
             self.bad_rows.append("Multiple possible committees for " + committee_key + ": " + committee_name)
 
-    def get_participating_committees(self, committee1, committee2, subcommittee1, subcommittee2):
+    def get_participating_committees(self, row):
```
beamalsky: Nice! Cool to see how you've redone this.

committeeoversightapp/management/commands/import_data.py
```diff
@@ -214,108 +303,163 @@ def get_participating_committees(self, committee1, committee2, subcommittee1, su
...
+            if committee.organization:
+                committee = committee.organization
+                new_committee, created = EventParticipant.objects.get_or_create(name=committee.name, event=event, organization=committee, entity_type=entity_type)
+            else:
+                new_committee, created = EventParticipant.objects.get_or_create(name=committee.lugar_name, event=event, entity_type=entity_type)
 
         except (ObjectDoesNotExist, AttributeError, ValueError) as e:
-            self.bad_rows.append("Row " + str(self.row_count) + ": Bad committee key " + committee_key)
+            self.bad_rows.append("Row " + str(self.row_index) + ": Bad committee key " + committee)
+        except MultipleObjectsReturned:
```
beamalsky: What case is this covering? Why is there no error logging?

committeeoversightapp/management/commands/import_data.py
```diff
@@ -6,13 +6,16 @@
...
-from django.db import IntegrityError
+from django.db import transaction, connection
 from django.db.models import Q
 
+
 from opencivicdata.core.models import Organization, OrganizationName
 from opencivicdata.legislative.models import Event, EventSource, EventParticipant
 from committeeoversightapp.models import HearingCategory, Committee
 
+ScrapedEvents = Event.objects.exclude(sources__note='spreadsheet')
```
beamalsky: At this point, is it possible that we'd be running the import script once Lugar has entered hearings using the web form? I hope not, but if so I think we'd need to check that imported hearings aren't in that queryset as well.

committeeoversightapp/management/commands/import_data.py
fgregg: okay, I can fix this. there were some problems with the makefile and the source row was not correct for me,  so I didn't undrerstand this value.

committeeoversightapp/management/commands/import_data.py
fgregg: i don't think I understand this. Could you expand.

beamalsky: Here's how the web form saves sources:

`source = EventSource(event=event, note="web form")`

So, instead of ScrapedEvents you could have something like ExistingEvents that also includes hearings with `(sources__note='web form')`

beamalsky: Actually never mind, sorry! I misread the query and didn't realize it was just excluding spreadsheet imports. Still, the name ScrapedEvents would be misleading if the DB has anything entered from the web form and I'd rename this to ExistingEvents or something similar.

fgregg:
I've resolved these concerns. I'm going to go ahead and bring this in

PR #64, opened 2019-09-05
Set up pytest and PR template

beamalsky:
This PR sets up pytest, at last, and a simple `hearing` fixture for later use. It also adds a PR template based on the [`just spaces` PR template](https://github.com/datamade/just-spaces/blob/master/.github/pull_request_template.md).

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,14 @@
+## Overview
+
+Closes #XXX
+
+Brief description of what this PR does, and why it is needed.
+
+
+## Testing Instructions
+
+ * How to test this PR
+ * Prefer bulleted description
+ * Start after checking out this branch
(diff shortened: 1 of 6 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,14 @@
+## Overview
```
hancush: 👋

tests/conftest.py
```diff
@@ -0,0 +1,38 @@
+import pytest
+from committeeoversight.wsgi import *
```
hancush: hm, this feels a little unorthodox. can you set `DJANGO_SETTINGS_MODULE=committeeoversight.settings` in `setup.cfg` instead?

beamalsky: Done!

beamalsky:
Oh yes good call! That sure is the point here. Will do and merge.

PR #73, opened 2019-09-10
Category detail page

### LINKED ISSUE #67: Category detail pages
Make category detail pages. These pages should include:

1. Category name
2. Filtered table of hearings assigned the category
3. Category description

Much like #66, I'm hoping we can find a way to make Django DetailViews, or a similar concept, play nicely with Wagtail Pages. I suspect this will involve overriding the Page's get_context_data method. If we can, then 1 and 2 will be drawn from the hearings data and 3 will be editable in the Wagtail backend.

beamalsky:
## Overview

Connects #67. This PR creates a Wagtail Page for automatically populated and dynamically updated Category detail pages. No CSS work yet but the elements are all there! I've left in-line notes below.

## Testing Instructions

* Delete your existing volume (not positive you'll have to do this, but I did): 
```
docker-compose down
docker volume rm committee-oversight_committeeoversight-db-data
``` 
 * Restart the app: 
```
docker-compose up
```
 * Open a new terminal tab. Load in the sample data:
```
docker-compose exec postgres pg_restore -C -j4 --no-owner -U postgres -d hearings /app/hearings.dump
```
 * Apply new migrations: 
```
docker-compose run --rm app python manage.py migrate
```
 * Navigate to http://localhost:8000/cms/pages/2/ and click "ADD CHILD PAGE"
*  Select "Category detail page"
* Create a new category page and make sure everything's intuitive. Save it as a draft; publish it; add images and text to the "body" field. Delete it. 
* Try different categories and make sure the hearings table is populating correctly (the hearings in `hearings.dump` don't have categories; you'll have to go to http://localhost:8000/list to edit them)


committeeoversightapp/models.py
```diff
@@ -9,17 +19,88 @@ class HearingCategoryType(models.Model):
...
+                                  null=True,
+                                  blank=True,
+                                  primary_key=False)
+    organization = models.ForeignKey(Organization,
+                                     null=True,
+                                     blank=True,
+                                     on_delete=models.CASCADE)
+
+
+class StaticPage(Page):
```
beamalsky: This is modeled on the LISC basic page template: https://github.com/datamade/lisc-cnda/blob/master/backend/cms/models.py#L10

This will be finished in a separate PR but was useful to get the Category pages working smoothly in the CMS backend.

Wagtail [StreamFields](https://docs.wagtail.io/en/v2.6.1/topics/streamfield.html) are very neat and allow the CMS user to put in an arbitrary number of different elements you define (in this case: headers, text blocks, images, and buttons) in whatever order they'd like. The `Page` model this inherits from also automatically provides a `title` attribute, so I didn't have to name it here but it will show up in the editing interface.

committeeoversightapp/models.py
```diff
@@ -9,17 +19,88 @@ class HearingCategoryType(models.Model):
...
+         icon='site'))
+    ])
+
+    # Editor configuration
+    content_panels = [
+        FieldPanel('category'),
+        StreamFieldPanel('body'),
+    ]
+
+    def save(self, *args, **kwargs):
```
beamalsky: Overriding the default `Page` `save()` function here to make sure the page title and draft title match the category selected when the page is saved

committeeoversightapp/static/committeeoversightapp/js/category_slug.js
```diff
@@ -0,0 +1,9 @@
+$(function() {
```
beamalsky: We need Javascript and a [Wagtail hook](https://docs.wagtail.io/en/v2.6.1/reference/hooks.html) to dynamically update the page's slug and SEO title 

The method is based on this blog post, which I've also dropped into a comment in `models.py`: https://timonweb.com/tutorials/how-to-hide-and-auto-populate-title-field-of-a-page-in-wagtail-cms/

committeeoversightapp/templates/committeeoversightapp/category_detail_page.html
```diff
@@ -0,0 +1,27 @@
...
+  $(document).ready(function () {
+      var id = {{page.category.id}}
+
+      $('#hearing-table').DataTable({
+        "dom": "Bfrtip",
+        "order": [[ 0, "desc" ]],
+        "columnDefs": [{ "orderable": false, "targets": [3, 4] }],
+        "processing": true,
+        "serverSide": true,
+        "ajax": "{% url 'order_list_json' 'category' %}/" + id
```
beamalsky: Grab the category ID number and append `category/<id>` to the url serving the JSON for serverside DataTables loading

committeeoversightapp/templates/list.html
```diff
@@ -6,19 +6,8 @@
...
-      <th scope="col">Hearing Date</th>
-      <th scope="col">Edit</th>
-      <th scope="col">Delete</th>
-    </tr>
-  </thead>
-  <tbody>
-  </tbody>
-  </table>
+
+ {% include 'partials/hearing_table.html' %}
```
beamalsky: Since we're using this in many parts of the site, I broke it out into a partial

committeeoversightapp/views.py
```diff
@@ -82,7 +84,21 @@ class EventListJson(BaseDatatableView):
     max_display_length = 500
 
     def filter_queryset(self, qs):
-        # based on example at https://pypi.org/project/django-datatables-view/
+        # in order to filter by categories and committees, grab the detail
+        # object type and its pk from the url and filter
```
beamalsky: This is the wonkiest part of this PR, imo: in order to filter hearings by category or committee, we need to pass this `EventListJson` view that context. The way I found to do that was to modify the `order_list_json` url to accept a variable, and then pass both the type of filter and object ID in the template.

The wonky part is that this view is then parsing the URL path into two parts, the `detail_type` and `pk`, in order to filter the table. Curious what you think here @hancush, and if you see a way to do this without URL parsing!

hancush:
@beamalsky – looking good! i'm having some trouble with the testing instructions. specifically:

1. i needed to set a slug to save my category page - is that expected? 
2. i'm able to view the hearings listing at http://localhost:8000/hearing/list, but i can't edit any of the hearings – i get 404s, e.g., at http://localhost:8000/edit/ocd-event/9a85634c-5fe1-462f-b875-3cac4c7b4b9e/.

hancush:
i'm going to have a look at the general approach anyway!

committeeoversightapp/models.py
```diff
@@ -9,17 +19,108 @@ class HearingCategoryType(models.Model):
...
+         icon='site'))
+    ])
+
+    # Editor configuration
+    content_panels = Page.content_panels + [
+        StreamFieldPanel('body'),
+    ]
+
+
+class LandingPage(Page):
```
hancush: what is the difference between a static page and a landing page?

beamalsky:
@hancush I fixed the broken `edit` and `delete` links on the `hearing/list` page, thanks for pointing that out! You shouldn't need to edit the slug manually—could you show me what's happening?

committeeoversightapp/models.py
```diff
@@ -9,17 +19,108 @@ class HearingCategoryType(models.Model):
...
+    # Editor configuration
+    content_panels = Page.content_panels + [
+        StreamFieldPanel('body'),
+    ]
+
+
+
+class CategoryDetailPage(Page):
+    # model page method adapted from https://timonweb.com/tutorials/how-to-hide-and-auto-populate-title-field-of-a-page-in-wagtail-cms/
+    category = models.ForeignKey(
```
hancush: right now, it's possible to try and create a new category page for a category that's already in use. you only get an error at the point of trying to save the redundant page, when the slug is autogenerated. it's a bit jarring, but then i may be imagining workflows that no one will actually do irl. do you think any action is needed to prevent creating redundant category pages, e.g., limiting the initial category choices?

committeeoversightapp/templates/committeeoversightapp/category_detail_page.html
```diff
@@ -0,0 +1,27 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <div class="col-12 col-lg-9 mx-auto">
+    <h1 class="mb-4">Category {{page.category.id}}: {{page.category.name}}</h1>
+  </div>
+
+  {% include 'partials/streamfield.html'%}
+
+  {% include 'partials/hearing_table.html' %}
```
hancush: this is rendering funny for me.

Screen Shot 2019-09-11 at 3 36 09 PM


i think your top-level element in the stream field partial may need to be a col element, but i'm not sure!

committeeoversightapp/views.py
hancush: ooh! i had to solve a similar problem in pensions. i ended up [adding url parameters to my ajax request, with values from my template context](https://github.com/datamade/bga-pensions/blob/0179475b15baeda43749afeff59967e0d6d45186/templates/index.html#L293), then [using them to filter the queryset in the view](https://github.com/datamade/bga-pensions/blob/0179475b15baeda43749afeff59967e0d6d45186/pensions/views.py#L331). i think it's a little more explicit than parsing the url... but i'm open to your input, too! what do you think?

committeeoversightapp/views.py
```diff
@@ -82,7 +84,21 @@ class EventListJson(BaseDatatableView):
...
+        path = self.request.path
+        split_path = path.split('/')
```
hancush: total aside: i have a weird bugaboo about avoiding accessing elements by their index, unless it's absolutely necessary. unpacking into multiple variables is my fave way to get around it, e.g., `a, b = 'some/url'.split('/')`. not a required change, esp if you make other changes to this route, just some fun trivia.

committeeoversightapp/views.py
```diff
@@ -91,8 +107,8 @@ def filter_queryset(self, qs):
 
     def prepare_results(self, qs):
         json_data = []
-        edit_string = "<a href=\"/edit/{}\"><i class=\"fas fa fa-pencil-alt\" id=\"edit-icon\"></i></a>"
-        delete_string = "<a href=\"/delete/{}\"><i class=\"fas fa fa-times-circle\" id=\"delete-icon\"></i></a>"
+        edit_string = "<a href=\"edit/{}\"><i class=\"fas fa fa-pencil-alt\" id=\"edit-icon\"></i></a>"
```
hancush: the edit and delete buttons are working in the hearing listing, but not from the category detail view. if they should be visible in that view, you'll want to start your url with a slash (`/hearings/edit/`, iirc?) so the link urls aren't relative to the current request location. https://webmasters.stackexchange.com/questions/56840/what-is-the-purpose-of-leading-slash-in-html-urls

committeeoversightapp/models.py
beamalsky: The landing page needs a table with all the committees and their ratings, per the wireframes. Static pages correspond to about, methodology, and contact in the wireframes.

committeeoversightapp/models.py
beamalsky: I'm thinking about that as a nice-to-have, since it would take some time to implement, the site already prevents users from making redundant pages, and we're on a tight deadline. Is there a reason you think it's worth putting time into that now?

committeeoversightapp/templates/committeeoversightapp/category_detail_page.html
beamalsky: This PR is for getting all the right data on the page, not the design! There's lots to be done there.

committeeoversightapp/views.py
beamalsky: Ah true thank you!

committeeoversightapp/models.py
hancush: got it! is creating two models what allows us to define two separate templates for the different page types?

committeeoversightapp/models.py
hancush: nope, just wanted to confirm that it's the expected behavior!

hancush: agree with your rationale for not doing it, fwiw!

committeeoversightapp/views.py
beamalsky: Using url parameters is a much nicer approach! Switched it.

committeeoversightapp/models.py
beamalsky: Yep! Each model in Wagtail corresponds to a template.

hancush: right on! would you mind leaving a comment to the effect of "define static and landing pages as separate models in order to define separate templates" on one or both of these models? it wasn't immediately clear to me why we'd need what appeared to be a duplicate model.

PR #81, opened 2019-09-17
Add CommitteeRating model and import command

### LINKED ISSUE #77: Set up sample Committee rating data
Lugar is still working on their committee rating system, but we'll need sample data to show for October. To set this up we'll want a new CommitteeRating model with committee, rating, and congress attributes. Jamie is getting data to us this week with some sample data to populate it, which will be a fixture for now and eventually derived in some way from Lugar's analysis.

hancush:
## Overview

Closes #77.

This PR:

- Defines a new `CommitteeRating` models. 
- Adds a `load_committeeratings` management command that loads in a hard-coded sample of ratings. 

## Notes

I chose to write a management command, rather than to include a fixture, because it's easier to update a sample dictionary than it is to update a fixture. Moreover, the management command can be repurposed to import structured ratings data when it becomes available.

## Questions

- I used a dump of the staging database to test the management command because `CommitteeRating` is related to `Committee`, but the dump included in this repository doesn't have any `Committee` objects. Should the dump be updated? Alternatively, I could update `import_data` to accept an argument instructing it to import committees (creating the `Committee` objects), but not hearings (which takes a long time). What do you think? **`Committee` is a vestigial model from work to sync up Lugar's spreadsheets with the scraped data. The ratings model should be associated with the OCD `Organization` model, which is in the dump.**
- Two questions about the new `Congress` model:
    - The new model would facilitate queries like `Congress.committeerating_set.all()` to get all ratings for a particular congress. Is this worth making queries more complicated? E.g., we'd want to select related objects when querying for ratings to avoid multiple DB calls. Happy to revert congress back to an attribute of the `CommitteeRating` model, if we only need to query by committee, not congress. **Factored this out. We don't want to query in this way.**
    - Open Civic Data [includes a legislative session model](https://github.com/opencivicdata/python-opencivicdata/blob/master/opencivicdata/legislative/models/session.py) that we could use instead of a custom `Congress` model. While this would allow us to lean yet further into the OCD standard, it requires start and end dates. Congressional start dates are inconsistent year over year, so we'd need to source and import these. This feels more like an enhancement to me, but I wanted to clear the choice to eschew this model – for now – with you!  **Tabling for now.**

## Testing Instructions

 * Stop your application containers and remove your Postgis volume: `docker-compose down && docker volume rm committee-oversight_committeeoversight-db-data`.
 * Restart your application containers and load in data as directed in the README. Apply the new migration: `docker-compose exec app python manage.py migrate`
 * Run the new management command: `docker-compose exec app python manage.py load_committeeratings`
 * Confirm the output looks like:
```
Added B rating for Senate Committee on Veterans' Affairs in 116th Congress
Added D rating for Senate Committee on Veterans' Affairs in 115th Congress
Added A rating for Senate Committee on Veterans' Affairs in 114th Congress
Added C rating for Senate Committee on Veterans' Affairs in 113th Congress
Added C- rating for Senate Committee on Veterans' Affairs in 112th Congress
Added A rating for Senate Committee on Veterans' Affairs in 111th Congress
Added C rating for Senate Committee on Veterans' Affairs in 110th Congress
Added B rating for Senate Committee on Veterans' Affairs in 109th Congress
Added F rating for Senate Committee on Veterans' Affairs in 108th Congress
Added F rating for Senate Committee on Veterans' Affairs in 107th Congress
Added C rating for Senate Committee on Energy and Natural Resources in 116th Congress
Added F rating for Senate Committee on Energy and Natural Resources in 115th Congress
Added C rating for Senate Committee on Energy and Natural Resources in 114th Congress
Added B+ rating for Senate Committee on Energy and Natural Resources in 113th Congress
Added D- rating for Senate Committee on Energy and Natural Resources in 112th Congress
Added A rating for Senate Committee on Energy and Natural Resources in 111th Congress
Added D rating for Senate Committee on Energy and Natural Resources in 110th Congress
Added B rating for Senate Committee on Energy and Natural Resources in 109th Congress
Added C rating for Senate Committee on Energy and Natural Resources in 108th Congress
Added F rating for Senate Committee on Energy and Natural Resources in 107th Congress
```


beamalsky:
Your thinking for the Congress model and using an import command makes sense to me! 

Committees are stored as OCD `Organizations` in the table `opencivicdata_organization`. Instead of a new `Committee` model, we should be able to use that `Organization` model as a foreign key directly for the `committee` attribute in `Committee` ratings. Let me know if you have trouble finding those committees or if you'd like to discuss.

committeeoversightapp/models.py
```diff
@@ -52,6 +53,19 @@ class Committee(models.Model):
...
+    committee = models.ForeignKey(Organization, on_delete=models.CASCADE)
+    congress = models.IntegerField()
+    rating = models.CharField(max_length=100)
+
+    @property
+    def congress_label(self):
+        '''
+        116 => '116th Congress'
+        '''
+        return '{} Congress'.format(ordinal(self.congress))
```
beamalsky: Cool!

committeeoversightapp/models.py
```diff
@@ -52,6 +53,19 @@ class Committee(models.Model):
                                      on_delete=models.CASCADE)
 
 
+class CommitteeRating(models.Model):
+    committee = models.ForeignKey(Organization, on_delete=models.CASCADE)
```
hancush: @beamalsky right now, a committee can have multiple ratings for a single congress. is this ok? if not, i can add a unique constraint, so we don't end up with nonsensical data.

hancush:
@beamalsky this is ready for re-review! i left a comment on the models file for your input.

committeeoversightapp/models.py
beamalsky: A rating should be unique! I'm now afraid that Lugar is going to be asking for multiple types of ratings for each committee for each Congress, so it makes sense to me to hang off on that change in anticipation of having to add a "rating type" column or something similar to this model. Up to you though!

hancush: Great instinct to hold off until we know more. I'll wait on it!

PR #82, opened 2019-09-18
Landing page

beamalsky:
## Overview

Closes #82. This PR creates a new landing page and visual style for the site. It should look like this:

Screen Shot 2019-09-18 at 4 42 21 PM

When a committee is selected from the search box, the site should redirect to that committee page.

This PR also includes an updated `hearings.dump` file.

## Testing Instructions

 * Stop your application containers and remove your Postgis volume: `docker-compose down && docker volume rm committee-oversight_committeeoversight-db-data`
* Restart your application: `docker-compose up`
* In another tab, load in data from `hearings.dump` (this will take a minute): `docker-compose exec postgres pg_restore -C -j4 --no-owner -U postgres -d hearings /app/hearings.dump`
* Navigate to http://localhost:8000/accounts/login/, login, play around, and make sure it looks like the above
* Use the search bar to find the House Committee on Science, Space, and Technology. Go to the page and make sure it shows you a Sample Committee Page


committeeoversightapp/templates/committeeoversightapp/landing_page.html
```diff
@@ -1,10 +1,82 @@
...
+{% load wagtailcore_tags %}
+{% load committeeoversight_tags %}
 
 {% block content %}
-  <div class="col-12 col-lg-9 mx-auto">
-    <h1 class="mb-4">{{page.title}}</h1>
+
+  <div class="row">
+    <div class="col-12 col-lg-4 col-xl-4 landing-description">
+        <h1 class="mb-4 ml-3 d-minus-rating">{{page.title}}</h1>
```
hancush: this is a very minor detail, but for clarity's sake, would you mind defining a separate class for the header color? it could be `header-color`, if this is a common header color across pages, or something more generic like `orange-color`. i feel bad classifying our site title as d minus! 😂

committeeoversightapp/templates/delete.html
```diff
@@ -13,7 +13,7 @@ <h1>Delete Hearing Record</h1>
  <form method="POST" class="form">
    {% csrf_token %}
    <div class="float-right">
-     <a href="/list" class="btn btn-back">
+     <a href="{% url 'list-event' %}" class="btn btn-back">
```
hancush: gotta love the `url` function. small but mighty refactor!

committeeoversightapp/templates/committeeoversightapp/landing_page.html
```diff
@@ -1,10 +1,82 @@
...
+        <div class="m-3">{{page.body|richtext}}</div>
+        <label class="m-3">
+          <h4 class="mb-3">Search by committee:</h4>
+          <select id="committee-select" class="form-control select2-single">
+            {% for committee in committees %}
+              <option value="{{committee.extras|lookup:"cleaned_id"}}">{{committee.name}}</option>
```
hancush: i believe you can get at dictionary keys in a template using `.` notation, e.g., `dict.key`.

committeeoversightapp/templatetags/committeeoversight_tags.py
```diff
@@ -0,0 +1,32 @@
...
+
+register = template.Library()
+
+
+@register.filter(name='lookup')
+def lookup(value, arg):
+    return value[arg]
+
+
+rating_colors = {
```
hancush: i'd expect a value that's derived from a model attribute to be defined as a property on the model. what do you say?

committeeoversightapp/models.py
```diff
@@ -72,23 +72,40 @@ class StaticPage(Page):
...
     # Editor configuration
     content_panels = Page.content_panels + [
-        StreamFieldPanel('body'),
+        FieldPanel('body'),
     ]
 
+    def get_context(self, request):
+        context = super(LandingPage, self).get_context(request)
+        context['house_committees'] = Organization.objects.all().filter(
+                                    classification='committee'
```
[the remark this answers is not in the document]
beamalsky: True!

committeeoversightapp/models.py
```diff
@@ -72,23 +72,40 @@ class StaticPage(Page):
...
+                                    parent_id__name__in=['United States Senate']
+                                )
+
+        committees = context['house_committees'] | context['senate_committees']
+
+        # Since Wagtail strips slashes out of urls but OCD prefixes their orgs
+        # with 'ocd-organization/', we need to do some cleaning. The cleaned_id
+        # of each committee should be of the form 'committee-<committee #>'
+        for committee in committees:
+            if not committee.extras.get('cleaned_id'):
```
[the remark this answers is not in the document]
beamalsky: It could be derived from the committee IDs, which are already persistent—do you think this would be better as a template tag?

committeeoversightapp/templates/committeeoversightapp/landing_page.html
beamalsky: Ha very fair

committeeoversightapp/templates/delete.html
beamalsky: You love to see it !

committeeoversightapp/templates/committeeoversightapp/landing_page.html
beamalsky: This thread said no but I now realize it's 7 years old! https://stackoverflow.com/questions/8000022/django-template-how-to-look-up-a-dictionary-value-with-a-variable

I'll wait and see how https://github.com/datamade/committee-oversight/pull/82#discussion_r326280287 shakes out before making any changes here.

committeeoversightapp/templatetags/committeeoversight_tags.py
beamalsky: Interesting! It feels strange to attach a CSS name directly on a model, but I hadn't thought about this option. Would it be a property on the `CommitteeRating` model? The `Organization` model is defined in OCD so I wouldn't want to modify that.

committeeoversightapp/fixtures/initial_cms_content.json
```diff
@@ -1,293 +1,5 @@
...
-  "pk": 1,
-  "fields": {
-    "path": "0001",
-    "depth": 1,
-    "numchild": 0,
-    "name": "Root"
-  }
-}
-]
+
```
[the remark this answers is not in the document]
beamalsky: Ee Dockerization tripped this up! Thanks for catching. I'm investigating.

committeeoversightapp/templatetags/committeeoversight_tags.py
hancush: yes, on the `ComitteeRating` model. and to be clear, i mean a `@property`, like a method on the model, not like a model attribute that's stored in the database. e.g., https://github.com/datamade/committee-oversight/blob/8ab986b2cee90e83fd925483d8d0727b5ee08d73/committeeoversightapp/models.py#L61-L66

beamalsky: Gotcha. Makes sense! I'll switch it over once I'll rebased this from master and have that model.

committeeoversightapp/fixtures/initial_cms_content.json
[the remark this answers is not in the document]
beamalsky: This is a broken piece of the sample data import process that didn't break here and isn't holding anything up (Wagtail data is in the current `hearings.dump`), but should be fixed soon. Would you be ok with me opening a separate issue and handling this outside of the current PR @hancush?

committeeoversightapp/models.py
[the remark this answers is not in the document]
hancush: i don't think a template tag would necessarily be appropriate, because we will want to use `cleaned_id` to route requests (i.e., get right object) to the committee detail page, in addition to rendering urls in the template. but i do think it makes sense to put this code somewhere it's easy to access and reusable. 

similarly to https://github.com/datamade/committee-oversight/pull/82#discussion_r326286522, i generally define values derived from a model's attributes as properties on the implicated model. do you think that could work here?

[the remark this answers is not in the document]
beamalsky: Wouldn't that have to be on the `Organization` model, which is imported by OCD?

committeeoversightapp/fixtures/initial_cms_content.json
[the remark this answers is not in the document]
hancush: i'd be more comfortable reverting this change, so broken code doesn't get back into master. you can check out the version of this file from the master branch like: `git checkout master -- committeeoversightapp/fixtures/initial_cms_content.json`

committeeoversightapp/models.py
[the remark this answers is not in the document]
hancush: adding a property would change the python api for the model, but it need not change the model at the database level. luckily, django offers [proxy models](https://docs.djangoproject.com/en/2.2/topics/db/models/#proxy-models) for this exact use case.

committeeoversightapp/fixtures/initial_cms_content.json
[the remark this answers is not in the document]
beamalsky: Looking into this a little deeper, I realized the problem isn't from Dockerization like I thought, but with an incorrect command last time I ran `dumpdata`! I'll fix it here.

committeeoversightapp/models.py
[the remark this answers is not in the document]
beamalsky: Wow interesting!! This is exactly what we need.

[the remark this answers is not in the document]
hancush: one thing to note, though, is that you can't use properties to filter querysets, so you can't do something like `Organization.objects.get(cleaned_id=blah)`, since `get` performs a database query, and we aren't storing `cleaned_id`. instead, you have to do the filtering in python. this can be ok – if your data isn't too big. about how many organizations will we be dealing with?

beamalsky:
@hancush this is ready for your eyes again! I think, once this is set, the next step is to rebase from master and pull in the CommitteeRatings you made. Does that make sense to you?

committeeoversightapp/models.py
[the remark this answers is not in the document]
beamalsky: Ah sorry missed this comment. There are around 40 organizations, so I think this will be fine.

committeeoversightapp/templatetags/committeeoversight_tags.py
```diff
@@ -0,0 +1,32 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter(name='lookup')
```
hancush: are you using this anyplace else? if not, i think it's safe to remove!

beamalsky: yes!

beamalsky:
@hancush I've now incorporated your sample committee ratings into the landing page. It's testable by following the testing instructions above, then running: `docker-compose run --rm app python manage.py load_committeeratings`

You should see something like this:

Screen Shot 2019-09-19 at 3 57 55 PM

Thanks for bearing with me through this messy PR!

committeeoversightapp/models.py
```diff
@@ -72,23 +140,33 @@ class StaticPage(Page):
...
+                                )
+        context['senate_committees'] = CommitteeOrganization.objects.all().filter(
+                                    classification='committee',
+                                    parent_id__name__in=['United States Senate']
+                                )
+
+        committees = context['house_committees'] | context['senate_committees']
+        context['committees'] = committees.order_by('name')
+
+        print("helskdjlfksgj")
```
hancush: `helskdjlfksgj`!!!!

beamalsky: ugh you're learning all my advanced testing secrets

PR #79, opened 2019-09-12
Add crontask to back up database

hancush:
## Overview

Closes #63.

This pull request adds a crontask to back up the database once per ~week~ day. It follows the database backup strategy [outlined in our deployment documentation](https://github.com/datamade/deploy-a-site/blob/master/Backups.md) (internal link).

## Testing Instructions

* Shell into the staging server and run `pg_dump -Fc -U postgres -d hearings | /usr/bin/aws s3 cp - s3://datamade-postgresql-backups/hearings/$(date -d "today" +"%Y%m%d%H%M").dump`.
* Navigate to the `datamade-postgresql-backups` bucket in the S3 console and confirm that a new backup was made in the `hearings` directory.
* Check that I spelled `committee-oversight-crontasks` right in the update to `after_install`. 😅

PR #88, opened 2019-09-24
Escape percent characters in cron tab

hancush:
## Overview

Related to #79.

Cron treats all percent characters as newlines. This PR escapes all of them, in the hopes that this solves our cron woes.


hancush:
More: http://www.pantz.org/software/cron/croninfo.html

PR #90, opened 2019-09-24
Add committee detail page

### LINKED ISSUE #66: Committee Detail pages
Make committee detail pages. This is somewhat subject to change during our wireframes work, but each committee page should show:

1. Committee Name
2. Filtered table of hearings that committee has held
3. Ratings for each Congress
4. Committee description
5. Potentially more metadata, like Chairperson, to be clarified with Lugar

I'm hoping we can find a way to make Django [DetailViews](https://docs.djangoproject.com/en/2.2/ref/class-based-views/generic-display/#detailview), or a similar concept, play nicely with Wagtail Pages. I suspect this will involve overriding the Page's `get_context_data` method. If we can, then 1 and 2 will be drawn from the `hearings` data and 3-5 will be editable in the Wagtail backend.

hancush:
## Overview

Closes #66 

This PR:

- Refactors common detail page functionality into an abstract `DetailPage` base class, and uses this base class to add a `CommitteeDetailPage`.
- Adds a `committee_detail_page.html` template.
- Enables, but hides, the Django admin interfaces for `CommitteeOrganization` and `CommitteeDetailPages` to enable autocomplete and take advantage of the autocomplete widget on the `CommitteeDetailPage` add and edit screens.
- Overrides the Javascript blocks in the edit and create Wagtail templates to put vendor Javascript before Wagtail Javascript. This allows us to hook into Django's Select2 Javascript, which was otherwise impossible, in order to dynamically set the slug and title on instances of `CommitteeDetailPage`.
- Adds a starter template for committee detail pages.

## Testing Instructions

- With the app running, run the migrations: `docker-compose exec app python manage.py migrate`
- In the Wagtail admin, navigate to Pages > The Committee Oversight Data Project.
- Click Add child page, and select Committee detail page.
- Publish a committee page or two. Confirm that the slug and title are updated as expected in the Promote tab.
- Back in the Pages listing, click the "Live" link to the far right to view your new detail pages. Confirm that the starter template contains the correct information and looks nice (enough).


committeeoversightapp/models.py
```diff
@@ -187,14 +197,51 @@ class CategoryDetailPage(Page):
...
 
-    # Editor configuration
+    def get_url_parts(self, *args, **kwargs):
+        '''
+        Route detail pages by their slug.
+        https://docs.wagtail.io/en/v2.6.1/topics/pages.html#obtaining-urls-for-page-instances
+        '''
+        return ('', '', '/' + self.slug)
+
+    def save(self, *args, **kwargs):
```
beamalsky: Beautiful :')

committeeoversightapp/models.py
```diff
@@ -187,14 +197,51 @@ class CategoryDetailPage(Page):
          icon='site'))
     ])
 
-    # Editor configuration
+    def get_url_parts(self, *args, **kwargs):
```
beamalsky: Where would this be used?

committeeoversightapp/views.py
```diff
@@ -92,7 +92,7 @@ def filter_queryset(self, qs):
         if detail_type == 'category':
             qs = qs.filter(hearingcategory__category_id=id)
         if detail_type == 'committee':
-            pass
+            qs = qs.filter(participants__organization_id=id)
```
beamalsky: We'll want hearings from a committee and also its subcommittees, so something like this:

```python
from django.db.models import Q

qs = qs.filter(Q(participants__organization_id=id) | Q(participants__parent_id=id))
```

committeeoversightapp/admin.py
```diff
@@ -1,3 +1,32 @@
...
+        '''
+        Override to exclude from admin interface:
+        https://stackoverflow.com/a/49295987/7142170
+        '''
+        return False
+
+
+class CommitteeOrganizationAdmin(HiddenAdminMixin, admin.ModelAdmin):
+    '''
+    Define this admin in order to take advantage of the Select2 mixin in the
```
beamalsky: Very clever!

committeeoversightapp/models.py
hancush: This customizes the routes for the detail pages. https://docs.wagtail.io/en/v2.6.1/topics/pages.html?highlight=get_url_parts#customising-url-patterns-for-a-page-model

hancush:
@beamalsky – responded to your comments! you'll want to zap your database and start afresh, since i changed the migration directly.

hancush:
@beamalsky what initial committee detail pages should the dump include? (or is it arbitrary?)

beamalsky:
It's arbitrary!

hancush:
@beamalsky updated, and also added the command to run the migrations after you do the initial db restore.

Issue #97, opened 2019-10-15
Add footer


beamalsky:
Add a footer, as included in the Moqups: https://app.moqups.com/NUARyAts8L/

This could be editable through the Wagtail admin as a snippet: http://docs.wagtail.io/en/v2.1.1/topics/snippets.html

beamalsky:
Closed by #99

PR #99, opened 2019-10-15
Update landing page and add footer

beamalsky:
## Overview

Closes #82 and #97. This PR updates the landing page styles and structure to match our [new mockups](https://app.moqups.com/NUARyAts8L/view/page/aa9df7b72). It also introduces a footer. Note that all committee data is dummy! It should look like this:

screencapture-localhost-8000-2019-10-16-11_54_27

## Testing Instructions

* Delete your volume (may not be necessary): `docker-compose down --volumes`
* Start the site back up: `docker-compose up`
* Open a new window and load `hearings.dump`:  
`docker-compose exec postgres pg_restore -C -j4 --no-owner -U postgres -d hearings /app/hearings.dump`
* Run migrations: `docker-compose exec app python manage.py migrate`
* Load CMS content: `docker-compose run --rm app python manage.py load_cms_content`
* Load sample committee ratings: `docker-compose run --rm app python manage.py load_committeeratings`
* View `http://localhost:8000` and make sure it looks like the screenshot above
* Make sure it looks good on mobile and at various sizes!


committeeoversightapp/models.py
```diff
@@ -51,6 +51,19 @@ def latest_rating(self):
         rating_set = self.committeerating_set.order_by('-congress')
         return rating_set[0]
 
+
+    def sliced_name(self):
```
hancush: Something like `display_name` would be a little more straightforward. It might also be nice to define this as a property (adding the `@property` decorator), since it's more of a derived attribute of an instance than functionality in and of itself! What do you think?

committeeoversightapp/models.py
```diff
@@ -51,6 +51,19 @@ def latest_rating(self):
...
+        chambers = {
```
hancush: This is very clever, but I'm finding it a little hard to follow! Any reason a simpler approach, e.g., [`string.replace`](https://docs.python.org/3.6/library/stdtypes.html#str.replace) or [`re.sub`](https://docs.python.org/3.6/library/re.html#re.sub), wouldn't work?

committeeoversightapp/templates/committeeoversightapp/landing_page.html
```diff
@@ -74,6 +43,19 @@ <h3 class="mb-4 text-right">The Senate</h3>
...
+
+    $(document).ready(function () {
+        $('.committee-table').DataTable({
+          "dom": "Bfrtip",
+          "paging": false,
+          "searching": false,
+          "info": false,
+          "order": [[ 0, "asc" ]],
+          "columnDefs": [
+            { "orderable": false, "targets": [3, 4, 5] },
```
hancush: Sorting by letter grade yields A, A+, A-, ..., F, F+, F-. Do you want to open an issue to address this later? If it helps, I tweaked the `natural` sort algorithm (https://github.com/datamade/risk-and-reach/pull/155/files#diff-1bf986eed1cc230d18c0075b6800592a) to sort mixtures of numbers and strings in a DataTable. There may be a lighter weight option for this use case!

committeeoversightapp/models.py
beamalsky: I think we're going to be using the full (unsliced) name for the committee detail pages, and this condensed name is just for tables where space is precious. Is there another name that makes more sense to you, with that in mind? Maybe `short_name`? I'll add the property decorator.

committeeoversightapp/models.py
beamalsky: Does switching out `split` for `replace` address your concern, or did you want to see a deeper change to the logic?

committeeoversightapp/templates/committeeoversightapp/landing_page.html
beamalsky: Great point! I'll open an issue to track this. It's possible these letters will just be letters, so we should hold off until Lugar decides.

committeeoversightapp/models.py
hancush: `short_name` sounds great!

committeeoversightapp/models.py
hancush: I mean the dictionary and iteration are a bit circuitous. You could achieve this more directly like

```python
if self.parent.name in ('United States House of Representatives', 'United States Senate'):
    return self.name.replace('House Committee on ', '').replace('Senate Committee on ', '')
```

or even

```python
# if parent.name...
return re.sub(r'(House|Senate) Committee on ', '', self.name)
```

PR #103, opened 2019-10-16
Add committee metadata

### LINKED ISSUE #101: Add metadata to Committee detail page
This corresponds with scope item 1.8:

**1.8. Administrative Interface**
The existing interface for entering and editing hearing details will form the basis of an administrative interface. We will also build facilities for editing any metadata that may be necessary for the individual committee pages, such as Chairperson, Ranking Member, and other members as well and the political parties of the members.

**8 hours | $1,200 USD**

Lugar wants two pieces of committee metadata: 

1. Chairperson
2. Whether or not to hide the CHP score

Ideally, we'd show both of these as inputs in the Wagtail committee editing interface. The tricky part here is to then grab that information and use it elsewhere on the site.

beamalsky:
## Overview

Closes #101. As described in the issue, this PR adds `chair` and `hide_rating` fields to the `CommitteeDetailPage` model for use elsewhere on the site. It also adds corresponding properties to the `CommitteeOrganization` model, so that this data is more easily accessible on the landing page.

## Testing Instructions

 * If you haven't already, follow the setup instructions in #99 
* Run migrations: `docker-compose exec app python manage.py migrate`
* Look at the landing page (http://localhost:8000/). There should be no chairs listed, and each committee should have a rating.
* Go into the Wagtail admin and create a new committee page. Give it a chair and select "hide rating"
* Go back to the landing page and make sure that the chair name shows up and the rating is "N/A"


beamalsky:
These are values Lugar staff should be able to edit, and from a user perspective it seems easiest to me to keep all CMS-editable content about a committee in one place.

hancush:
@beamalsky I hear you about convenience, though I think it might be counterintuitive to the user to change the display of a committee on the landing page by editing its detail page. On the data side, I think it would be more appropriate for those values to be attributes of the Committee. 

What do you think about creating a simple committee metadata editor in the CMS? It seems like Wagtail provides facilities for this: http://docs.wagtail.io/en/latest/reference/contrib/modeladmin/

beamalsky:
We chatted IRL, and I pushed for keeping `CommitteeOrganization` a pure proxy model for OCD's `Organization` and homing user-editable attributes on the `CommitteeDetailPage`. 

@hancush agreed we could stick with the following principle: In order to keep a single editable Wagtail page for each model and maintain separation between scraped and editable data, user-entered and -editable data should live as attributes on a model's DetailPage; scraped or OCD-sourced data should be attached to the corresponding base model. 

I added a comment to that effect. @hancush do you want to take another look at this? Or let me know if you still have misgivings about this approach.

PR #111, opened 2019-10-29
Category detail pages

beamalsky:
## Overview

Closes #67 and #108. This PR adds a simple category detail page following the mockup styles: https://app.moqups.com/NUARyAts8L/view/page/ad113671c

It also adds two `db-ops` services to simplify the local testing process.

## Testing Instructions

 * Take down your Docker container: `docker-compose down --volumes`
*  Load data: `docker-compose -f docker-compose.yml -f docker-compose.db-ops.yml run -e PGPASSWORD=postgres --rm dbload-dump`
* Load fixtures: `docker-compose -f docker-compose.yml -f docker-compose.db-ops.yml run --rm dbload-fixtures`
* Run the site: `docker-compose up`
* Log into the site and create a new Category Detail Page
* Check that the title, and URL populate automatically as you'd expect
* Publish the page and look at it live. Make sure all looks good!
* Edit a hearing to give it the category whose page you made, and make sure it appears in the hearings table

PR #112, opened 2019-11-06
Hearing detail and list pages

### LINKED ISSUE #110: Hearing detail pages
Following: https://app.moqups.com/NUARyAts8L/view/page/a5a349309

beamalsky:
## Overview

Closes #110. This PR creates templates for hearing detail pages and a Wagtail hearing list page. It reorganizes the hearing editing/creating/deleting pages to accommodate this phase of the project, and creates links between the Committee, Hearing, and Category pages wherever possible.

This PR also updates the README to reflect changes to the dbload service made in #111.

**Note:** All hearings should now have detail pages, but Lugar will create Committee and Category detail pages manually as they enter their descriptions of both. The CMS fixture in this PR includes a sample page of each: the [House Committee on Science, Space, and Technology](http://localhost:8000/committee-2c6395b2-dc9d-45f6-a4db-481b1b750431/) and the category page for [Nominations](http://localhost:8000/category-nominations/)

## Testing Instructions

 * Follow the instructions in the [README from this PR](https://github.com/datamade/committee-oversight/tree/issue-110-hearing-details) to get the site running
* Open http://localhost:8000/ and click `Browse Hearings` in the navbar. Make sure you see a list of the categories and a table of hearings
* Click `Nominations` in the category list and make sure the link works
* Go back to the `Browse Hearings` page and click on a particular hearing. Does it look as you'd expect?
* Log into the site and click around. Can you edit and create pages as you'd expect? Are the pages linked as you'd expect them to be?


committeeoversightapp/templates/committeeoversightapp/category_detail_page.html
```diff
@@ -21,7 +21,7 @@ <h2 class="mb-4 text-center">Category: {{page.category}}</h2>
 
   <div class="card mt-5 col-12 col-xl-10 mx-auto card-shadow">
     <div class="card-body">
-      {% include 'partials/hearing_table.html' %}
+      {% include 'partials/hearing_table.html' with table_size="table-sm" %}
```
beamalsky: The hearings table on the Browse Hearings page doesn't need to be as compact as it is on the Committee and Category detail pages.

committeeoversightapp/templates/hearing_delete.html
```diff
@@ -0,0 +1,31 @@
+{% extends 'base.html' %}
```
beamalsky: This file was just renamed `hearing_delete.html` from `delete.html`. The same applies to `hearing_edit.html` and `hearing_create.html`

committeeoversightapp/utils.py
```diff
@@ -4,28 +4,36 @@
 
 from django.core.exceptions import ObjectDoesNotExist
 
-from opencivicdata.legislative.models import EventDocument, EventDocumentLink, \
```
beamalsky: This file is unchanged other than flake8 fixes!

committeeoversightapp/templatetags/committeeoversight_tags.py
```diff
@@ -0,0 +1,32 @@
+from django import template
+from slugify import slugify
+
+from committeeoversightapp.models import CommitteeOrganization
+
+register = template.Library()
+
+
+@register.filter
+def get_category_url(category):
```
jeancochrane: A custom filter works perfectly fine for this function (and the rest of the functions in this module), but I often implement operations like this one (i.e. performs some processing on a single object, returns a value instead of e.g. rendered HTML) as methods or `@property` methods on the relevant model class, since it feels more object-oriented and often makes it easier to test.

Not a required change, just an alternate possibility I wanted to point out.

committeeoversightapp/templates/hearing_detail.html
```diff
@@ -0,0 +1,116 @@
...
+            {% endfor %}
+         </td>
+       </tr>
+       <tr>
+         <td><b>Category</b></td>
+         <td>
+          {% if category|get_category_url %}
+            <a href="{{ category|get_category_url }}">{{category}}</a>
+          {% else %}
+            {{category}}
```
jeancochrane: Seems like `category` can be `None`, which will display the string `None` on the page, so we'll want to give this a fallback:

```suggestion
            {{category|default_if_none:""}}
```

Screen Shot 2019-11-06 at 5 04 35 PM

committeeoversightapp/templates/hearing_detail.html
```diff
@@ -0,0 +1,116 @@
...
+     <tbody>
+       <tr>
+         <td><b>Hearing Date</b></td>
+         <td>{{hearing.start_date}}</td>
+       </tr>
+       <tr>
+         <td><b>Committees</b></td>
+         <td>
+           {% for committee in committees %}
+             {% if committee|is_subcommittee %}
```
jeancochrane: A similar thought here about template tags vs. model methods.

committeeoversightapp/templatetags/committeeoversight_tags.py
jeancochrane: I'm guessing maybe that your reasoning here was that these models are OCD models instead of committeeoversight models -- if that's the case, a [proxy model](https://docs.djangoproject.com/en/2.2/topics/db/models/#proxy-models) might be able to give you the access you need to the model class.

committeeoversightapp/templates/hearing_detail.html
```diff
@@ -0,0 +1,116 @@
...
+    <h2 class="text-center">{{hearing.name|title }}</h2>
+    <br />
+
+    <table class="table table-striped table-bordered" id="hearing-table">
+     <thead>
+     </thead>
+     <tbody>
+       <tr>
+         <td><b>Hearing Date</b></td>
+         <td>{{hearing.start_date}}</td>
```
jeancochrane: Since `start_date` is stored as a raw string, the output isn't very friendly:

Screen Shot 2019-11-06 at 5 08 27 PM

Not a required change I don't think, but if it bothers you then you could think about doing some parsing to try to format the date.

committeeoversightapp/urls.py
```diff
@@ -1,11 +1,11 @@
 from django.conf.urls import url, include
 from committeeoversightapp.views import EventCreate, EventList, EventDelete, \
-                                        EventEdit, EventListJson
+                                        EventEdit, EventListJson, EventDetail
 
 urlpatterns = [
     url(r'accounts/', include('django.contrib.auth.urls')),
     url(r'^hearing/create/$', EventCreate.as_view(), name='create-event'),
-    url(r'^hearing/list/$', EventList.as_view(), name='list-event'),
+    url(r'^hearing/view/(?P<pk>.*?/.*?)/', EventDetail.as_view(), name="detail-event"),
```
jeancochrane: I find these kinds of regexes hard to grok. Is there a reason you prefer `url()` with a raw regex to [`path()`](https://docs.djangoproject.com/en/2.2/ref/urls/#path) with converters?

committeeoversightapp/utils.py
```diff
@@ -4,28 +4,36 @@
...
 def archive_url(url):
     wayback_host = 'http://web.archive.org'
     save_url = '{0}/save/{1}'.format(wayback_host, url)
     archived = requests.get(save_url)
     try:
-        archive_url = '{0}{1}'.format(wayback_host, archived.headers['Content-Location'])
+        archive_url = '{0}{1}'.format(
+            wayback_host,
+            archived.headers['Content-Location']
+            )
```
jeancochrane: Nitpick, but it seems like this and other function calls/dictionary definitions that you've updated in this module have inconsistent indentation for the closing paren. Sometimes it's on a new line indented with the previous arg, and sometimes it's on the same line as the last arg in the list.

If the closing paren is on its own line, I usually match it to the opening line, e.g.:

 ```python
        archive_url = '{0}{1}'.format(
            wayback_host,
            archived.headers['Content-Location']
        )
```

Since we don't have a styleguide this is ultimately a personal preference thing though, as long as it's consistent I think it's fine.

File: committeeoversightapp/views.py [hunk 10]
```diff
@@ -104,14 +105,15 @@ def filter_queryset(self, qs):
 
     def prepare_results(self, qs):
         json_data = []
+        detail_string = "<a href=\"/hearing/view/{0}\">{1}</a>"
```
jeancochrane: What's your thinking here behind hardcoding the link instead of using `reverse()` to generate it on the fly?

committeeoversightapp/views.py
```diff
@@ -120,18 +122,45 @@ def prepare_results(self, qs):
...
+        #get category context
+        try:
+            context['category'] = HearingCategory.objects.get(event=context['hearing']).category
+        except ObjectDoesNotExist:
+            context['category'] = None
+
+        #get committee context
+        # context['committees'] = context['hearing'].participants.filter(entity_type="organization")
+        # print("hello")
+        # print(context['committees'])
```
jeancochrane: Some commented-out code here that can probably be deleted.

committeeoversightapp/views.py
```diff
@@ -249,7 +278,7 @@ def post(self, request, **kwargs):
             save_documents(event, transcript_form.cleaned_data)
             save_witnesses(event, witness_formset.cleaned_data)
 
-        return redirect('list-event')
+        return redirect('/hearings')
```
jeancochrane: Similarly, should this be a hardcoded path or a reference to a view name?

committeeoversightapp/templatetags/committeeoversight_tags.py
```diff
@@ -0,0 +1,32 @@
+from django import template
+from slugify import slugify
```
jeancochrane: As we talked about today, you could eliminate a dependency here if you used Djangos' built in [`slugify()` function](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.text.slugify):

```suggestion
from django.utils.text import slugify
```

requirements.txt
```diff
@@ -1,6 +1,7 @@
 django>=2.1.5,<2.2
 pytest==5.1.2
 pytest-django==3.5.1
+python-slugify==4.0.0
```
jeancochrane: Remember to get rid of this dependency if you decide to go with `django.utils.text.slugify`.

committeeoversightapp/management/commands/load_cms_content.py
```diff
@@ -21,14 +21,19 @@ def handle(self, **options):
...
+        try:
+            Site.objects.all().delete()
+            Page.objects.all().delete()
+            PageRevision.objects.all().delete()
+            Image.objects.all().delete()
+            LandingPage.objects.all().delete()
+            StaticPage.objects.all().delete()
+            CategoryDetailPage.objects.all().delete()
+            CommitteeDetailPage.objects.all().delete()
+            HearingListPage.objects.all().delete()
```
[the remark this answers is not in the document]
beamalsky: Oo nice. This is deleting some but not all models in the `committeeoversightapp`, `wagtailcore`, and `wagtailimages` apps. Is there a different method you'd suggest here?

committeeoversightapp/templates/hearing_detail.html
beamalsky: This formatting is site-wide and I'm not sure what Lugar's preferences are, so I'd prefer to handle it outside of this issue!

committeeoversightapp/templates/hearing_detail.html
beamalsky: I'm intentionally displaying None here. All hearings should have categories, so I wanted to make clear something is missing.

committeeoversightapp/views.py
beamalsky: I'll use reverse!

committeeoversightapp/views.py
beamalsky: The `/hearings` page is now created in Wagtail, since we need the intro text to be editable, so it doesn't have a view slug.

committeeoversightapp/urls.py
beamalsky: This is just the convention I started following... some of the OCD urls are wonky, since their IDs includes slashes, but let me see if I can refactor.

beamalsky: Ok! I see that `url` is slated for deprecation, so I've converted all urls in this project to `path` or `re_path`. I don't think it's worth taking more time to figure out how to do this without regex.

beamalsky:
Thanks for this feedback @jeancochrane! This is ready for your eyes again.

committeeoversightapp/management/commands/load_cms_content.py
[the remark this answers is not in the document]
jeancochrane: Ah, that's a tough one -- I didn't realize it was a selection of models instead of the full set.

One approach that would be more object-oriented would be to define an optional class attribute, something like `Model.reset_on_load`, that determines whether or not the model should be cleared when loading data. Then you could set this attribute on models you control, and update the reset code to be something like:

```python
# Do this after deleting Wagtail models, since you don't have control over them
for Model in apps.get_app_config('committeeoversightapp').get_models():
    if Model.reset_on_load:
        Model.objects.all().delete()
```

However, this doesn't really reduce the maintenance burden introduced by this functionality, it simply moves it from the management command to the models themselves. I think this is probably a better place for it to be, since a developer adding a new model down the road is more likely to copy what they see in `models.py` than to notice they need to update a management command, but it's ultimately a matter of taste.

committeeoversightapp/templates/hearing_detail.html
jeancochrane: Sounds reasonable to me 👍

committeeoversightapp/templates/hearing_detail.html
jeancochrane: Ah, OK. Carry on then!

committeeoversightapp/views.py
jeancochrane: Oh interesting! Seems like an important limitation of Wagtail that I wasn't aware of.

committeeoversightapp/urls.py
jeancochrane: Seems fine to me, I just wanted to make sure you were aware of `path()` since it offers a simplified API.

committeeoversight/urls.py
```diff
@@ -25,9 +25,9 @@
 from committeeoversightapp.views import pong
 
 urlpatterns = [
-    path(r'^admin/', admin.site.urls),
-    path(r'^pong/$', pong,),
```
beamalsky: Got warnings for some lingering Regex I didn't remove when I switched this from `url()`. Fixed in this commit! (https://github.com/datamade/committee-oversight/commit/07d4aea32914092d3b75d1e7f2981eca010c3383)

committeeoversightapp/models.py
```diff
@@ -185,7 +185,12 @@ def __str__(self):
         return self.rating
 
 
-class StaticPage(Page):
+class ResetMixin(object):
```
beamalsky: @jeancochrane what do you think of this mixin approach?

jeancochrane: Really nice 👍

committeeoversight/urls.py
jeancochrane: Good catch!

PR #115, opened 2019-11-11
Add permanent committee list & fix JS slugify

### LINKED ISSUE #113: Limit landing page committees
Per Jamie's instructions, limit landing page to the following current permanent committees: 

**House**

House Committee on Agriculture

House Committee on Appropriations

House Committee on Armed Services

House Committee on Budget

House Committee on Education

House Committee on Energy and Commerce

House Committee on Financial Services

House Committee on Foreign Affairs

House Committee on Homeland Security

House Committee on House Administration

House Committee on Intelligence (Permanent Select)

House Committee on Judiciary

House Committee on Natural Resources

House Committee on Oversight and Government Reform

House Committee on Rules

House Committee on Science, Space, and Technology

House Committee on Small Business

House Committee on Transportation and Infrastructure

House Committee on Veterans' Affairs

House Committee on Ways and Means

 

**Senate**

Senate Committee on Aging

Senate Committee on Agriculture, Nutrition, and Forestry

Senate Committee on Appropriations

Senate Committee on Armed Services

Senate Committee on Banking, Housing, and Urban Affairs

Senate Committee on Budget

Senate Committee on Commerce, Science, and Transportation

Senate Committee on Energy and Natural Resources

Senate Committee on Environment and Public Works

Senate Committee on Finance

Senate Committee on Foreign Relations

Senate Committee on Health, Education, Labor, and Pensions

Senate Committee on Homeland Security and Governmental Affairs

Senate Committee on Indian Affairs

Senate Committee on Intelligence

Senate Committee on Judiciary

Senate Committee on Rules and Administration

Senate Committee on Small Business and Entrepreneurship

Senate Committee on Veterans' Affairs

beamalsky:
## Overview

Closes #113. We need to display only the current permanent committees on the landing page, a list provided by Lugar. This PR adds that list as `CURRENT_PERMANENT_COMMITTEES`, a settings variable. It also adds `CHAMBERS` as another handy global variable.

In the process of putting together this PR, I realized that Django's slugify function and the JS I'm using don't exactly match up, specifically in their handling of the category "Meeting/Markup". I've corrected that inconsistency here as well.

There's one outstanding question that I've sent to Jamie and Dan: 

> This update is just about ready to push to the staging site, but I have a question about one committee: in the current data, we have a 'House Committee on Education and the Workforce.' Is that ok to sub in for 'House Committee on Education', or should we look into changing the data?

Question for @jeancochrane: what do you think about putting these things in `settings`? Is there a different place that you'd recommend?

## Testing Instructions

* Follow the instructions in the README to get the site running
* Make sure the landing page works and shows exactly 39 committees

### DIFF
--- a/committeeoversight/settings.py
+++ b/committeeoversight/settings.py
@@ -140,3 +140,50 @@ MEDIA_URL = '/media/'
 LOGIN_REDIRECT_URL = '/'
 
 WAGTAIL_SITE_NAME = 'Committee Oversight'
+
+CURRENT_PERMANENT_COMMITTEES = [
+    'House Committee on Agriculture',
+    'House Committee on Appropriations',
+    'House Committee on Armed Services',
+    'House Committee on Budget',
+    'House Committee on Education and the Workforce',
+    'House Committee on Energy and Commerce',
+    'House Committee on Financial Services',
+    'House Committee on Foreign Affairs',
+    'House Committee on Homeland Security',
+    'House Committee on House Administration',
+    'House Committee on Intelligence (Permanent Select)',
+    'House Committee on Judiciary',
+    'House Committee on Natural Resources',
+    'House Committee on Oversight and Government Reform',
+    'House Committee on Rules',
+    'House Committee on Science, Space, and Technology',
+    'House Committee on Small Business',
+    'House Committee on Transportation and Infrastructure',
+    'House Committee on Veterans\' Affairs',
+    'House Committee on Ways and Means',
+    'Senate Committee on Aging',
+    'Senate Committee on Agriculture, Nutrition, and Forestry',
+    'Senate Committee on Appropriations',
+    'Senate Committee on Armed Services',
+    'Senate Committee on Banking, Housing, and Urban Affairs',
+    'Senate Committee on Budget',
+    'Senate Committee on Commerce, Science, and Transportation',
+    'Senate Committee on Energy and Natural Resources',
+    'Senate Committee on Environment and Public Works',
+    'Senate Committee on Finance',
+    'Senate Committee on Foreign Relations',
+    'Senate Committee on Health, Education, Labor, and Pensions',
+    'Senate Committee on Homeland Security and Governmental Affairs',
+    'Senate Committee on Indian Affairs',
+    'Senate Committee on Intelligence',
+    'Senate Committee on Judiciary',
+    'Senate Committee on Rules and Administration',
+    'Senate Committee on Small Business and Entrepreneurship',
+    'Senate Committee on Veterans\' Affairs',
+]
+
+CHAMBERS = [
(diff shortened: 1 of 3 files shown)


committeeoversight/settings.py
```diff
@@ -140,3 +140,50 @@
...
+    'Senate Committee on Foreign Relations',
+    'Senate Committee on Health, Education, Labor, and Pensions',
+    'Senate Committee on Homeland Security and Governmental Affairs',
+    'Senate Committee on Indian Affairs',
+    'Senate Committee on Intelligence',
+    'Senate Committee on Judiciary',
+    'Senate Committee on Rules and Administration',
+    'Senate Committee on Small Business and Entrepreneurship',
+    'Senate Committee on Veterans\' Affairs',
+]
```
jeancochrane: This is a problem I run into all the time and still don't have a great answer for -- how should you organize "global variables" that should apply across an entire Django project? I used to store them in the settings file like this, and I still think that works fine for most use cases.

However, lately I've moved in the direction of storing them in a dedicated app-level module, as in https://github.com/datamade/just-spaces/blob/a4f109d1aeb69884103eb0b4cb248576983476fa/fobi_custom/plugins/form_elements/fields/types.py. My thinking here is that `settings.py` is intended for project-level configuration values, typically for configuring app behavior. Sometimes global variables are customizing an app, but more often they actually represent static data that is relevant to a specific app, and not necessarily the whole project (where a project is an instantiation of a given app).

Our line between project/app is usually pretty fuzzy, since most of our apps are single-use apps meant to represent a single site. So for the most part I don't think this decision matters much. The one exception is probably if you're using a separate settings file for tests, in which case these values should definitely be in an app-level module to prevent duplication -- but using separate settings files for tests is an anti-pattern that I hope we can move away from anyway.

committeeoversightapp/static/js/detail_editor_slug.js
```diff
@@ -34,18 +34,13 @@ try {
...
+  // taken from https://gist.github.com/mathewbyrne/1280286
+  // to match Django slugify
+  function slugify(text) {
+    return text.toString().toLowerCase()
+      .replace(/\s+/g, '-')           // Replace spaces with -
+      .replace(/[^\w\-]+/g, '')       // Remove all non-word chars
+      .replace(/\-\-+/g, '-')         // Replace multiple - with single -
+      .replace(/^-+/, '')             // Trim - from start of text
+      .replace(/-+$/, '');            // Trim - from end of text
+  }
```
jeancochrane: Good catch 👍

committeeoversight/settings.py
beamalsky: Makes sense, very helpful to hear your thinking on this! I think I'm going to leave these in settings but may change that if there's more global static data to handle in the future.

PR #117, opened 2019-11-12
CHP Scores

### LINKED ISSUE #102: CHP score data handling
Though we don't yet have their exact methodology, we know that Lugar's CHP scores will be calculated from counts of different types of hearings from the current and past Congress. We could calculate these all dynamically on page load, but in the interest of a fast site I propose that we write an import script that will calculate CHP scores and their constituent parts and be run nightly.

I'm imagining the following new classes:

`CommitteeRating`:
- `id` - auto assigned
- `committee` - foreign key to CommitteeOrganization (exists already)
- `congress` - foreign key to Congress (see below)
- `investigative_oversight_hearings` - integer
- `policy_legislative_hearings` - integer
- `total_hearings` - integer
- `chp_points` - int

methods
- `chp_score` - method; %
- `max_chp_points` - method; int
- `investigative_oversight_avg` - method; int
- `policy_legislative_avg` - method; int
- `total_avg` - method; int

`Congress`:
- `id` - number of the Congress
- `start_date`
- `end_date`

beamalsky:
## Overview

Closes #102. This PR follows the method described in #102 to implement a new `load_committeeratings` with the CHP score methodology, then use a variety of methods on the `CommitteeOrganization` and `CommitteeRatings` models to populate the data on the site. The testing instructions below include directions to download a data dump with (limited) categorization data entered by Lugar. 

There are a bunch of calculations going on here that can get confusing with the partial data we have—I didn't realize how few hearings have categorizations on the staging site, which I'm going to bring up with Lugar next week. For now, I wanted to pause here to get feedback on this approach.

## Testing Instructions

 * We need hearings data with categories to test the ratings system, so we can't keep using the sample data previously in `hearings.dump`. Go to the S3 bucket with backups from the hearings site, download the most recent (where I've deleted the committeeratings), and save it in this project's root as a new `hearings.dump`: https://s3.console.aws.amazon.com/s3/object/datamade-postgresql-backups/hearings/201911131730.dump?region=us-east-1&tab=overview
 Since this file is so large, I'm not keeping it under version control.
* Run `docker-compose down --volumes`
* Follow setup instructions in the README
* Go to http://localhost:8000/ and view the CHP scores you've just calculated. Verify that there are hearing counts attached to surveys.
* Visit the page for the [House Committee on Science, Space, and Technology](http://localhost:8000/committee-61caba6d-71bc-41c7-882e-32ec97768f63) and make sure the detail page is populating as you'd expect
* Visit the two comparison pages linked from the landing page


jeancochrane:
Since we won't make any progress on this PR until Monday at the earliest, how would you feel about me passing off review responsibility of this PR to @hancush? The approach as a whole looks good to me as we talked about on Tuesday, but it'd be great to get Hannah's take before moving further ahead.

beamalsky:
Sounds good to me! @hancush and I can check in at our sprint planning this morning.

committeeoversightapp/models.py
```diff
@@ -94,10 +88,208 @@ def chair(self):
     def hide_rating(self):
         return CommitteeDetailPage.objects.get(committee=self.id).hide_rating
 
+    @property
+    def ratings(self):
+        return self.committeerating_set.all().order_by('-congress')
+
+    @property
+    def ratings_reverse(self):
```
hancush: `ratings` / `ratings_reverse` imply that the default ratings ordering will be used, but you implement a custom sort in the properties. Is there a more descriptive way to name these properties, so ordering is self-evident? Maybe `ratings_in_congress_order` or something to that effect?

committeeoversightapp/models.py
```diff
@@ -94,10 +88,208 @@ def chair(self):
...
+class CommitteeRating(models.Model):
+    committee = models.ForeignKey(CommitteeOrganization, on_delete=models.CASCADE)
+    congress = models.ForeignKey(Congress, on_delete=models.CASCADE)
+    investigative_oversight_hearings = models.IntegerField(null=True, blank=True)
+    policy_legislative_hearings = models.IntegerField(null=True, blank=True)
+    total_hearings = models.IntegerField(null=True, blank=True)
+    chp_points = models.IntegerField(null=True, blank=True)
+
+    @property
+    def chp_score(self):
```
hancush: Really neat to see this tie all the other properties together. Very clever!

committeeoversightapp/models.py
```diff
@@ -300,6 +450,39 @@ class CommitteeDetailPage(DetailPage):
         FieldPanel('hide_rating'),
     ]
 
+    def get_context(self, request):
+        context = super(CommitteeDetailPage, self).get_context(request)
+        context['latest_rating'] = context['page'].committee.latest_rating
+
+        congresses = context['page'].committee.ratings_reverse \
+            .values_list('congress_id', flat=True)
+        context['congresses'] = [x for x in congresses]
```
hancush: Picking nits over here: Do you need to unpack this into a list, or will the template work with the output from `values_list`?

hancush: Ditto for each of the similar lines below!

committeeoversightapp/management/commands/load_cms_content.py
```diff
@@ -7,6 +7,7 @@
 from django.apps import apps
 from django.core.management.base import BaseCommand
 from django.core.management import call_command
+from django.core.exceptions import ObjectDoesNotExist
```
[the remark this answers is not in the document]
beamalsky: To catch this error in line 34, yep!

committeeoversightapp/management/commands/load_committeeratings.py
```diff
@@ -1,100 +1,66 @@
...
+            Q(participants__organization__parent_id=committee.id)
+        ).filter(start_date__range=(
+            congress.start_date,
+            congress.end_date
+        ))
+
+        # Investigative Oversight = Agency Conduct Hearings + Private Sector Hearings
+        investigative_oversight_hearings=self.count_by_category(
+            committee_hearings,
+            [4, 5]
```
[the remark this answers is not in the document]
beamalsky: Both IDs and category names should be stable, so I'll use names for better readability

committeeoversightapp/models.py
```diff
@@ -94,10 +88,208 @@ def chair(self):
     def hide_rating(self):
         return CommitteeDetailPage.objects.get(committee=self.id).hide_rating
 
+    @property
+    def ratings(self):
+        return self.committeerating_set.all().order_by('-congress')
```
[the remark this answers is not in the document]
beamalsky: Congresses should be ordered by `id`, which corresponds to their real-world number (ie, we're in the 116th Congress right now, which has an id of 116). Do you see a reason this wouldn't be consistent across environments? I've updated this to explicitly refer to `congress__id`

committeeoversightapp/models.py
```diff
@@ -94,10 +88,208 @@ def chair(self):
...
+        if self.start_date < date.today() < self.end_date:
+            return True
+        else:
+            return False
+
+    @property
+    def percent_passed(self):
+        days_in_session = 668
+        days_passed = (date.today() - self.start_date).days
+        return round(days_passed * 100.0 / days_in_session, 2)
```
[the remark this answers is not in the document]
beamalsky: It shouldn't ever be over 100! I'll clamp it. 

For your first question, I'm in the habit of putting `100.0` second because it converts `int`s to `floats`. For example:

```python
 >>> days_in_session = 668
>>> days_passed = 400
>>> days_passed / days_in_session * 100.0
0.0
>>> days_passed * 100.0 / days_in_session
59.880239520958085
```

Is there something else you'd recommend here?

committeeoversightapp/models.py
beamalsky: These need to be straight lists, since they're going into my charthelper as Javascript!

beamalsky:
Thanks for your feedback @hancush! This is ready for your eyes again, with just one outstanding question.

committeeoversightapp/models.py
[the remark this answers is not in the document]
hancush: I did not actually know this, but apparently Python 3.x will always do floating point division, regardless of the types of the inputs. https://codereview.stackexchange.com/a/69544

```bash
(committee-oversight) call-me-hank:committee-oversight hannah$ python3
Python 3.7.3 (default, Mar 27 2019, 09:23:15)
[Clang 10.0.1 (clang-1001.0.46.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 3/5
0.6
```

You can do the future import to safeguard against people running the app in Python 2.x, though chances are slim thanks to our explicit Python version in our Dockerfile!

committeeoversightapp/models.py
```diff
@@ -94,10 +88,215 @@ def chair(self):
     def hide_rating(self):
         return CommitteeDetailPage.objects.get(committee=self.id).hide_rating
 
+    @property
+    def ratings_by_congress_desc(self):
```
hancush: Perfect!

committeeoversightapp/models.py
```diff
@@ -94,10 +88,215 @@ def chair(self):
...
+        if self.start_date < date.today() < self.end_date:
+            return True
+        else:
+            return False
+
+    @property
+    def percent_passed(self):
+        days_in_session = 668
+        days_passed = (date.today() - self.start_date).days
+        percent_passed = round(days_passed * 100.0 / days_in_session, 2)
```
hancush: Given https://github.com/datamade/committee-oversight/pull/117#discussion_r349276996, what do you think about reordering?

committeeoversightapp/models.py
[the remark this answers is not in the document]
beamalsky: Huh TIL! Thanks for digging into this.

PR #130, opened 2019-12-11
Adjust hearings table data displayed

### LINKED ISSUE #128: Adjust hearing table displays
From Jamie:

- Hide hearings categorized as Meeting/Markup, Postponed, Report, Cancelled, and Testimony and exclude those categories on the “Browse Hearings” page
- Sort newest to oldest based on hearing date (`start_date`)
- Columns: DATE, HEARING TITLE, COMMITTEE, CATEGORY

### LINKED ISSUE #131: Hide login in footer
> Another addition that is only tangentially related to design is that the “login” button should probably not be an option that casual users should be able to see, as this function will only be used by Lugar Center staff and interns. Ideally, this login button could not be seen by the public at all. But assuming that it must be on the site somewhere, is there a place we can bury that button so that we can minimize the number of public viewers who can see it?

beamalsky:
## Overview

Closes #128.
Closes #131. 

This PR responds to a list of requested changes from Jamie detailed in #128 and #131.

## Testing Instructions

 * Run the site following directions in the README
* Visit http://localhost:8000/hearings/ and make sure you see only the categories listed in `settings.DISPLAY_CATEGORIES` in the list in the top
* Click through a few pages of hearings and make sure you don't see any hearings listed with other categories
* Log in through the footer and make sure that link works
* Go back to http://localhost:8000/hearings/ and make sure you see hearings from all categories
* Visit http://localhost:8000/hearings/, a category detail page, and a committee detail page and make sure the hearings appear newest to oldest by Hearing Date and show the columns  DATE, HEARING TITLE, COMMITTEE, CATEGORY (except for COMMITTEE on committee detail pages and CATEGORY on category detail pages)

### DIFF
--- a/committeeoversightapp/models.py
+++ b/committeeoversightapp/models.py
@@ -521,7 +521,9 @@ class HearingListPage(ResetMixin, Page):
 
     def get_context(self, request):
         context = super(HearingListPage, self).get_context(request)
-        context['categories'] = HearingCategoryType.objects.all()
+        context['categories'] = HearingCategoryType.objects.filter(
+            name__in=settings.DISPLAY_CATEGORIES
+        )
         return context
 
 
--- a/committeeoversightapp/models.py
+++ b/committeeoversightapp/models.py
@@ -9,6 +9,7 @@ from django.db import models
 from django.db.models import Max, Avg
 from django.db.models.fields import TextField, BooleanField
 from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
 
 from wagtail.core.models import Page
 from wagtail.core.fields import StreamField, RichTextField
@@ -22,6 +23,25 @@ from opencivicdata.core.models import Organization
 from opencivicdata.legislative.models import Event, EventParticipant, \
                                              EventDocument
 
+
+class HearingEvent(Event):
+    class Meta:
+        proxy = True
+
+    @property
+    def category(self):
+        try:
+            return HearingCategory.objects.get(
+                event_id=self.id
+            ).category
+        except ObjectDoesNotExist:
+            return None
+
+    @property
+    def committees(self):
+        return CommitteeOrganization.objects.filter(
+            eventparticipant__event=self.id
+        )
 
 class CommitteeManager(models.Manager):
     def permanent_committees(self):
@@ -45,6 +65,7 @@ class CommitteeManager(models.Manager):
         context['house_committees'] = self.house_committees()
         context['senate_committees'] = self.senate_committees()
         return context
+
 
 class CommitteeOrganization(Organization):
     class Meta:
--- a/committeeoversightapp/templates/partials/hearing_table.html
+++ b/committeeoversightapp/templates/partials/hearing_table.html
@@ -2,9 +2,10 @@
   <table class="table table-striped table-borderless {{ table_size }}" id="hearing-table">
      <thead>
        <tr>
-         <th scope="col">Date Updated</th>
-         <th scope="col" class="hearing-title-col">Hearing Title</th>
-         <th scope="col">Hearing Date</th>
+         <th scope="col">Date</th>
+         <th scope="col">Hearing Title</th>
+         <th scope="col">Committee</th>
+         <th scope="col">Category</th>
          {% if request.user.is_superuser %}
            <th scope="col">Edit</th>
            <th scope="col">Delete</th>
--- a/committeeoversightapp/templates/partials/hearing_table.html
+++ b/committeeoversightapp/templates/partials/hearing_table.html
@@ -2,10 +2,10 @@
   <table class="table table-striped table-borderless {{ table_size }}" id="hearing-table">
      <thead>
        <tr>
-         <th scope="col">Date</th>
-         <th scope="col">Hearing Title</th>
-         <th scope="col">Committee</th>
-         <th scope="col">Category</th>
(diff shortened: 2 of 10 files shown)


committeeoversightapp/models.py
```diff
@@ -23,6 +24,25 @@
...
+    class Meta:
+        proxy = True
+
+    @property
+    def category(self):
+        try:
+            return HearingCategory.objects.get(
+                event_id=self.id
+            ).category
+        except ObjectDoesNotExist:
```
hancush: 100% fine to import this exception for use as you've done here, just want to share some related trivia: You can access a `DoesNotExist` exception via any Django model, too, e.g., `HearingCategory.DoesNotExist`. Accomplishes the same thing, just minimally more convenient. 🙃

committeeoversightapp/views.py
```diff
@@ -135,46 +171,28 @@ def prepare_results(self, qs):
...
-                event=context['hearing']
-            ).category
-        except ObjectDoesNotExist:
-            context['category'] = None
-
-        context['committees'] = CommitteeOrganization.objects.filter(
-            eventparticipant__event=context['hearing'].id
-        )
-
+        context['category'] = context['hearing'].category
```
hancush: What's the reason for explicitly setting these in the context, instead of accessing them via your `hearing` object in the template?

committeeoversightapp/views.py
```diff
@@ -111,18 +121,44 @@ def prepare_results(self, qs):
...
-                    item.updated_at.strftime("%Y-%m-%d %I:%M%p %Z"),
-                    detail_string.format(
-                        escape(reverse_lazy(
-                            'detail-event',
-                            kwargs={'pk':item.pk}
-                        )),
-                        escape(item.name.title())
-                    ),
-                    item.start_date,
+        for item in qs:
```
hancush: Hm, the logic in this loop is getting pretty tough to follow. Can you think of a way to break this into methods or roll some of this conditional string formatting back up into model properties?

committeeoversightapp/views.py
```diff
@@ -74,10 +75,19 @@ class EventList(LoginRequiredMixin, TemplateView):
...
-    columns = ['updated_at', 'name', 'start_date', 'id', 'id']
+    model = HearingEvent
+
+    # Define the columns that will be returned. These need to match attributes
+    # of the model but most will be calculated later, so 'id' here is a
+    # placeholder.
+    columns = [
+        'start_date',
+        'name',
+        'id', # committees
```
hancush: Are you required to include a value here, or can it be an empty string?

committeeoversightapp/views.py
```diff
@@ -135,46 +171,28 @@ def prepare_results(self, qs):
...
-                        escape(item.name.title())
-                    ),
-                    item.start_date,
-                    '',
-                    '',
-                ])
+                ]
+                json_data.append(row_data)
+            # if not authenticated, only include display hearing categories 
+            elif not item.category or item.category.name in settings.DISPLAY_CATEGORIES:
```
hancush: This is a little confusing to me. If there is not an item category, or if the category is in display categories, add two empty values?

hancush: `category_string` will be an empty value if there is not a category, and it is added to the row unconditionally in L156. Will this introduce a redundant empty value?

committeeoversightapp/views.py
beamalsky: A vestigial structure. Ty for catching!

committeeoversightapp/views.py
beamalsky: This does work with an empty string. Good idea!

committeeoversightapp/views.py
beamalsky: Refactored! Thanks for the nudge.

committeeoversightapp/views.py
beamalsky: I think my refactor addresses this question. I also realized that I'm not actually sure if hearings without a category should be displayed to non-admin users. I opened a question issue for it (#134) and emailed Jamie. For now, hearings with no category will only show up for admin users.

beamalsky:
The hearing list view logic should be easier to follow now. Ready for your eyes Monday @hancush !

committeeoversightapp/templates/partials/hearing_table.html
```diff
@@ -2,9 +2,10 @@
   <table class="table table-striped table-borderless {{ table_size }}" id="hearing-table">
      <thead>
        <tr>
-         <th scope="col">Date Updated</th>
-         <th scope="col" class="hearing-title-col">Hearing Title</th>
-         <th scope="col">Hearing Date</th>
+         <th scope="col" style="min-width:10%;">Date</th>
```
hancush: TIL you should do this, instead of the `col-*` classes, in Bootstrap 4! Thank you!!!

beamalsky:
Thx for the catch @hancush! The error you were seeing was from an incorrect column value in the committee detail page DataTables settings, which I've fixed. 

In my last commit I also caught an error in how I was building the JSON for the server side processing. I noticed that for non-authenticated users, reordering the columns after the page initially loaded caused the table to have some disappearing rows, showing 2 or 3 entries instead of 10. I figured out that this is because I was doing my filtering for authenticated users in the `prepare_results` function, which it seems expects an output of the same length as the input `qs`. I moved this filtering to the `filter_queryset` function in that view, where it should have been all along.

PR #136, opened 2019-12-16
Committee link handling

### LINKED ISSUE #132: Avoid broken committee links
We currently show a link to all committee pages. We should either:

1. Confirm from Lugar that they'll be creating pages for all committees
2. Only make links if a committee page exists, eg if it's a permanent committee

beamalsky:
## Overview

Closes #132. This PR sets up committee links to work for a site with a full set of detail pages for permanent committees, not assuming that any other committee pages will exist. On hearing detail pages and tables of hearings, a committee will now be:

- Listed in form `<PARENT COMMITTEE>`, `<COMMITTEE>` if it's a subcommittee
- Linked to its committee detail page, if permanent

## Testing Instructions

* Run the site following the README
 * View the Browse Hearings page and a committee detail page and make sure they load as expected
* Log in as an admin and pick a random hearing to edit. Add a couple committees and subcommittees, including at least one that's not permanent (see `settings.py` for that list). For example, here's what I did:
Screen Shot 2019-12-16 at 4 17 15 PM
* Check various views for this hearing and make sure it shows up correctly

### DIFF
--- a/committeeoversightapp/models.py
+++ b/committeeoversightapp/models.py
@@ -85,8 +85,8 @@ class CommitteeOrganization(Organization):
         return '/committee-' + self.id.split('ocd-organization/').pop()
 
     @property
-    def parent_url(self):
-        return '/committee-' + self.parent.id.split('ocd-organization/').pop()
+    def parent_proxy(self):
+        return CommitteeOrganization.objects.get(id=self.parent.id)
 
     @property
     def short_name(self):
@@ -100,6 +100,47 @@ class CommitteeOrganization(Organization):
             return True
         else:
             return False
+
+    @property
+    def is_permanent(self):
+        if self in CommitteeOrganization.objects.permanent_committees():
+            print("True")
+            return True
+        else:
+            return False
+
+    @property
+    def get_linked_html(self):
+        if self.is_subcommittee:
+            if self.parent_proxy.is_permanent:
+                return '<a href=\"{0}\">{1}</a>, {2}'.format(
+                    self.parent_proxy.url,
+                    self.parent,
+                    self
+                )
+            else:
+                return '{0}, {1}'.format(self.parent, self)
+        else:
+            if self.is_permanent:
+                return '<a href="{0}">{1}</a>'.format(self.url, self)
+            else:
+                return self
+
+    @property
+    def get_linked_html_short(self):
+        if self.is_subcommittee:
+            if self.parent_proxy.is_permanent:
+                return '<a href=\"{0}\">{1}</a>'.format(
+                    self.parent_proxy.url,
+                    self.parent
+                )
+            else:
+                return str(self.parent)
+        else:
+            if self.is_permanent:
+                return '<a href="{0}">{1}</a>'.format(self.url, self)
+            else:
+                return str(self)
 
     @property
     def chair(self):
--- a/committeeoversightapp/models.py
+++ b/committeeoversightapp/models.py
@@ -103,8 +103,7 @@ class CommitteeOrganization(Organization):
 
     @property
     def is_permanent(self):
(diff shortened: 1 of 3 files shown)


committeeoversightapp/models.py
```diff
@@ -85,8 +85,8 @@ def url(self):
         return '/committee-' + self.id.split('ocd-organization/').pop()
 
     @property
-    def parent_url(self):
-        return '/committee-' + self.parent.id.split('ocd-organization/').pop()
+    def parent_proxy(self):
```
hancush: Ooh, this is venturing very close to the behavior we implemented in `django-councilmatic` with [`django-proxy-overrides`](https://github.com/datamade/django-proxy-overrides). More on that here: https://github.com/datamade/django-councilmatic/tree/2.5#proxying

Meanwhile, this is a smart solution for limited needs! Just be aware it will fire at least two SQL queries – one to get the parent, and the other to get the `CommitteeOrganization` object – so it won't scale as well.

committeeoversightapp/models.py
```diff
@@ -101,6 +101,47 @@ def is_subcommittee(self):
...
+                    self.parent,
+                    self
+                )
+            else:
+                return '{0}, {1}'.format(self.parent, self)
+        else:
+            if self.is_permanent:
+                return '<a href="{0}">{1}</a>'.format(self.url, self)
+            else:
+                return self
```
hancush: `self`, not `str(self)`?

committeeoversightapp/templates/hearing_detail.html
```diff
@@ -19,14 +19,11 @@ <h2 class="text-center">{{hearing.name|title }}</h2>
...
          <td>
            {% for committee in committees %}
-             {% if committee.is_subcommittee %}
-                <a href="{{ committee.parent_url }}">
-                  {{committee.parent}}</a>, {{committee}}
-              {% else %}
-                <a href="{{ committee.url }}">{{committee}}</a>
-             {% endif %}
+             {% autoescape off %}
+              {{ committee.get_linked_html }}
```
hancush: Is this the same as `{{ variable|safe }}`?

committeeoversightapp/models.py
```diff
@@ -101,6 +101,47 @@ def is_subcommittee(self):
         else:
             return False
 
+    @property
+    def is_permanent(self):
+        if self in CommitteeOrganization.objects.permanent_committees():
+            print("True")
```
hancush:
```suggestion
```

committeeoversightapp/models.py
```diff
@@ -101,6 +101,47 @@ def is_subcommittee(self):
         else:
             return False
 
+    @property
+    def is_permanent(self):
+        if self in CommitteeOrganization.objects.permanent_committees():
```
hancush: This check will evaluate the permanent committees queryset and do the memebrship check in Python. I think it's more efficient to do something like `if CommitteeOrganization.objects.permanent_committees().filter(slug=self.slug).exists()`, which will just return True/False.

committeeoversightapp/models.py
beamalsky: Ah very cool to see, thanks for the links! I'm going to leave this for now but if we find that we're using these parent properties more, or if the current ones are slowing down the site, I'll plan to switch it.

committeeoversightapp/models.py
beamalsky: Good catch. For future clarity I switched this and other similar occurrences to `self.name`

committeeoversightapp/templates/hearing_detail.html
beamalsky: Yep https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#safe

PR #147, opened 2020-01-15
Add last updated calculations

### LINKED ISSUE #125: What should the "last updated" value on the site refer to?
Discuss with Jamie & Dan. Some options:
1. The date of the last hearing imported 
2. The date of the last import
3. The date of the last category assignment

Connects to #109.

beamalsky:
## Overview

Closes #125. This PR adds calculations for the site's "Last updated" values, which show when hearings with a category were last updated on the site. More detail on that definition in the issue. 

While putting this together, I realized that the `load_committeeratings.py` script was doing some slight miscounting due to the duplicate hearings fixed in #145. This PR also fixes that problem through the creation and use of `CommitteeOrganization.hearings`.

## Testing Instructions

* Run `docker-compose down --volumes` and rebuild the site following directions in the README
* Check out the "Last updated" values on the landing page, comparison pages, and a committee detail page. It should look something like this (though not this exact value):

Screen Shot 2020-01-15 at 5 10 35 PM

* Log in and [add a new hearing](http://localhost:8000/hearing/create/) **with no category** and for an existing committee page 
* View that committee detail page and make sure the "Last updated" value hasn't changed
* Edit the hearing you just made and give it a category
* Go back to the committee detail page and make sure "Last updated" is now changed to the current time


committeeoversightapp/models.py
```diff
@@ -73,6 +82,21 @@ class Meta:
 
     objects = CommitteeManager()
 
+    @property
+    def hearings(self):
+        return HearingEvent.objects.all().filter(
```
hancush: I believe you can get at related events through an Organization's `eventparticipant_set`. Is the purpose here to return `HearingEvent` objects instead of `Event` objects?

beamalsky: Yes, having this be `HearingEvents` lets us use the `category` value in the hearing tables more easily.

beamalsky:
I opened up an issue for caching: https://github.com/datamade/committee-oversight/issues/148

Merging! Thanks Hannah.

Issue #141, opened 2019-12-17
Switch Lugar to using production site


beamalsky:
- [x] Set up crontasks to run there (Ie, reverse the change made in #140)
- [x] Deploy a tagged commit
- [x] Sync production site db to match what's on the staging site

To restore the production database from the latest staging backup, I followed these steps.

1. Created a new backup in [this folder](https://s3.console.aws.amazon.com/s3/buckets/datamade-postgresql-backups/hearings/?region=us-east-1&tab=overview) by running:

```
ssh ubuntu@committeeoversight.datamade.us
(pg_dump -Fc -U postgres -d hearings | /usr/bin/aws s3 cp - s3://datamade-postgresql-backups/hearings/$(date -d "today" +"\%Y\%m\%d\%H\%M").dump) && echo "staging backup $(date -d "today" +"\%Y\%m\%d\%H\%M").dump complete" >> /tmp/committee-oversight-crontasks.log 2>&1
```

2. Visited [the folder](https://s3.console.aws.amazon.com/s3/buckets/datamade-postgresql-backups/hearings/?region=us-east-1&tab=overview) and downloaded the newest file into my `committee-oversight` root project directory as `hearings.dump`

3. Move `hearings.dump` to the production server, SSH into the server, and restore the database:

```
scp ~/Desktop/committee-oversight/hearings.dump ubuntu@committeeoversight-production.datamade.us:/tmp
ssh ubuntu@committeeoversight-production.datamade.us
sudo su postgres
dropdb hearings
createdb hearings
pg_restore -C -j4 --no-owner -U postgres -d hearings /tmp/hearings.dump | psql
```

- [x] Delete crontask from staging site

```
ssh ubuntu@committeeoversight.datamade.us
sudo rm /etc/cron.d/committee-oversight-crontasks
```

beamalsky:
To restore the production database from the latest staging backup, I followed these steps.

1. Created a new backup in [this folder](https://s3.console.aws.amazon.com/s3/buckets/datamade-postgresql-backups/hearings/?region=us-east-1&tab=overview) by running:

```
ssh ubuntu@committeeoversight.datamade.us
[...]

PR #162, opened 2020-02-14
Add committee ratings crontask

### LINKED ISSUE #160: Set up cronjob to update Grades

beamalsky:
## Overview

Closes #160. This PR sets up a crontask to run the `load_committeeratings` management command every hour to update grade calculations. 

Currently this will run only for the staging site. That will be switched to production in #141 

@hancush I ssh-ed into the staging site and ran this command with the current deployment ID to make sure it works. You could do the same to test, or we could deploy this and monitor it there.

### DIFF
--- a/scripts/render_configs.py
+++ b/scripts/render_configs.py
@@ -19,9 +19,13 @@ if __name__ == "__main__":
   nginx_template_path = '/home/datamade/committee-oversight-{0}/configs/committee-oversight-{1}.conf.nginx'.format(
     deployment_id, deployment_group)
   nginx_outpath = '/etc/nginx/conf.d/committee-oversight.conf'
+
   supervisor_template_path = '/home/datamade/committee-oversight-{0}/configs/committee-oversight-{1}.conf.supervisor'.format(
     deployment_id, deployment_group)
   supervisor_outpath = '/etc/supervisor/conf.d/committee-oversight.conf'
+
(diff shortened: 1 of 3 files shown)


scripts/render_configs.py
```diff
@@ -36,8 +40,16 @@
     supervisor_rendered = supervisor_conf.render(
       {'deployment_id': deployment_id})
 
+  with open(crontask_template_path) as f:
```
hancush: I have never noticed before that this file has only 2 spaces! We should fix that at some point... I'll leave it to you whether you want to do it in this PR.

beamalsky:
The crontask file will only be copied into the right place on staging deployments unless I'm looking at this line wrong: https://github.com/datamade/committee-oversight/pull/162/files#diff-8c59a62af65566df3850ac9526e72150L59

I suppose this doesn't handle deleting the old crontask file. Since I'll only be switching this from staging to production once, hopefully, I'm inclined to just handle that manually as part of #141. What do you think?

scripts/render_configs.py
beamalsky: Fixed! The spacing is fine in the [template](https://github.com/datamade/deploy-a-site/blob/master/templates/django/%7B%7Bcookiecutter.directory_name%7D%7D/scripts/render_configs.py), so I'm not sure how that happened.

PR #177, opened 2020-04-03
Add display names for committees to Wagtail

### LINKED ISSUE #174: Change 2 committee names
From Jamie: 

> There are two committees that have changed their names in the 116th Congress, including the House Committee on Education and Labor and the House Committee on Oversight and Reform. We need to change the name that appears on the landing page and at the top of those two pages. If it makes it easier, we don’t have a strong desire to change the name for individual hearings, so long as the title of those committee pages read correctly. Additionally, it might simplify things down the line if we keep the names as follows:
House Committee on Education
House Committee on Oversight and Reform
This will eliminate the need to change the name back and forth as the majority party changes.

beamalsky:
## Overview

Closes #174. This PR adds a `display_name` attribute to the committee detail pages. As committee names can shift slightly over Congress, but we'll want to retain each committee as a stable record, this will allow the Lugar Center to adjust committee names through the CMS.

## Testing Instructions

 * Download [this recent dump](https://s3.console.aws.amazon.com/s3/object/datamade-postgresql-backups/hearings/%255C2020%255C04%255C03%255C17%255C42.dump?region=us-east-1) from S3 and save it in the root of your project as `hearings.dump`
* Run `docker-compose down --volumes`
* Run `docker-compose -f docker-compose.yml -f docker-compose.db-ops.yml run -e PGPASSWORD=postgres --rm dbload-dump`
* Run `docker-compose run --rm app python manage.py migrate`
* Run `docker-compose up`
* Log into the site and edit a committee detail page. Change the `display name` at the bottom and make sure it shows up on the site, including on the landing page and hearing detail pages.
* Create a new committee page (you might have to delete an existing one). Make sure that it saves successfully without a modified display name.


committeeoversightapp/models.py
```diff
@@ -498,13 +512,16 @@ def title_field(self):
 
     def save(self, *args, **kwargs):
         title = str(getattr(self, self.title_field))
+
+        if not title:
```
beamalsky, line 516: I feel like there's probably a more elegant way to handle there being two possible sources for the site `title`... open to ideas @hancush!

hancush: You could do 
```python
str(getattr(self, self.title_field) or getattr(self, self.fallback_title_field))
```

but I wouldn't do anything fancier than that.

beamalsky:
I checked in with Jamie on whether or not they'd like the ability to see past display names assigned to a committee. She said they don't need that, and that the site should only show search results for the current name of a committee. Considering that setting up aliases as a new model would be more complex to get working in Wagtail, I think it makes sense to keep the current approach if it sounds reasonable to you @hancush.

hancush:
Thanks for checking, @beamalsky! In that instance, merge away. 🚀

PR #184, opened 2020-04-23
Add Congresses to model admin and implement footnotes

### LINKED ISSUE #178: Set up a way to add footnotes to Congresses
With the shortened length on the 116th Congress (#175), we need a way to document that decision. Jamie also mentioned that there are a couple other Congresses they'd like to add footnotes for. Set up a CMS-compatible system for Lugar to write up footnotes for individual Congresses and have them appear across the site. As Congresses are a model, this might be a good way to try out Wagtail's ModelAdmin http://docs.wagtail.io/en/v2.0/reference/contrib/modeladmin/

I have the go-ahead from Jamie to spend 5-10 hours on this

beamalsky:
## Overview

Closes #178. This PR adds the ability to create footnotes for each Congress on the site, and adds them to the templates following Jamie's request below. Also, while adding Congress to the ModelAdmin, I've added a more user-friendly way to adjust the `inactive_days` value for each Congress (as discussed in #176)

From Jamie, here are the 3 places footnotes should show up:

1. Next to the congress on individual committee pages (footnote should be placed directly under chart)
unnamed

2. Next to the “Projected Grade” on individual committee pages (footnote should be placed directly under chart) 
unnamed (1)

3. Next to Current Congress (projected) on landing page (footnote should be at bottom of page)
unnamed (2)

## Testing Instructions

 * If you haven't recently, download an updated db from production [here](https://s3.console.aws.amazon.com/s3/object/datamade-postgresql-backups/hearings/%255C2020%255C04%255C23%255C16%255C30.dump?region=us-east-1). Add it to the root of your project as `hearings.dump`
* `docker-compose -f docker-compose.yml -f docker-compose.db-ops.yml run -e PGPASSWORD=postgres --rm dbload-dump`
* `docker-compose run --rm app python manage.py migrate`
* `docker-compose up`
* Log into Wagtail and go to the "Edit Congresses" section of the left sidebar. Add one and then multiple footnote values  and make sure they shows up on the site as described above


committeeoversightapp/wagtail_hooks.py
```diff
@@ -23,3 +27,17 @@ def global_admin_css():
         '<link rel="stylesheet" href="{}">',
         static("css/wagtail-overrides.css")
     )
+
+class CongressAdmin(ModelAdmin):
+    model = Congress
+    menu_label = 'Edit Congresses'  # ditch this to use verbose_name_plural from model
```
hancush, line 33: Do we need these comments?

committeeoversightapp/models.py
```diff
@@ -194,7 +195,17 @@ def ratings_by_congress_desc(self):
 
     @property
     def ratings_by_congress_asc(self):
-        return self.committeerating_set.all().order_by('congress__id')
+        committeeratings = self.committeerating_set.all().order_by('congress__id')
+
+        count = 1
+        for rating in committeeratings:
+            if rating.congress.has_footnote:
```
hancush, line 202: Clever. This feels a bit unconventional, but I'm having trouble thinking of an alternative.

committeeoversightapp/models.py
```diff
@@ -297,6 +321,13 @@ def percent_passed(self):
 
         return cap_100(percent_passed)
 
+    @property
```
hancush, line 324: Does this offer any advantage over `if congress.footnote`?

committeeoversightapp/wagtail_hooks.py
beamalsky: Nope!

committeeoversightapp/models.py
beamalsky: Yep I also spent a while trying to think of something that's more built into Django, but this is where I landed. If you think of another way later I'd be curious to hear it! I think this is ok for one instance on the site, but if we had this same problem for other types of data it would be worth abstracting somehow.

committeeoversightapp/models.py
beamalsky: Good point 🤦🏻‍♀️

beamalsky:
Thanks @hancush! I added footnotes to the Congress comparison page like you suggested—though Jamie didn't specifically request it, I think you're right that consistency necessitates it.

Issue #149, opened 2020-01-22
Make sure individual hearing pages are indexed


beamalsky:
Using wget spider mode

beamalsky:
I ran `wget --spider --recursive oversight-index.thelugarcenter.org`. It crawled the static, committee, and category pages but not the hearing detail pages.

beamalsky:
I've added sitemaps for the hearings and Wagtail pages in #192. I also registered the site on the [Google Search Console](https://search.google.com/search-console) via an analytics account and manually registered those sitemap files. It looks like `sitemap-hearings` is working but `sitemap-wagtail` has incurrent URLs still including `localhost`.

beamalsky:
In the production CMS I changed the Site `hostname` to `oversight-index.thelugarcenter.org` and the port to 443. I refreshed the sitemap and it worked!

beamalsky:
Plan: Check back on the coverage at https://search.google.com/search-console/index?resource_id=https%3A%2F%2Foversight-index.thelugarcenter.org%2F in a day or two to make sure Google has properly indexed this. If it has, we can close this issue.

PR #192, opened 2020-05-13
Add sitemaps and analytics

beamalsky:
## Overview

Connects #149. I found that `wget`'s spider mode missed the individual hearing pages, so I've created a sitemap for this app. It comes in two parts: 
1. Wagtail's default generated sitemap, created using [their documentation](https://docs.wagtail.io/en/stable/reference/contrib/sitemaps.html)
2. A separate sitemap for the hearings, created using [GenericSitemap](https://docs.djangoproject.com/en/3.0/ref/contrib/sitemaps/#shortcuts)

One snag here is that the sitemap generated for #2 (http://localhost:8000/sitemap-hearings.xml) is just a blob of links, not the structured XML I'd expect to see. I tried creating this sitemap using `django.contrib.sitemaps.Sitemap` but got the same results. I'm wondering if this is something Django does to keep large sitemaps to a reasonable file size, but wasn't able to find documentation. Does this look familiar, @hancush?

I've also added a Google Analytics tag in this PR, to facilitate [pinging Google
](https://docs.djangoproject.com/en/3.0/ref/contrib/sitemaps/#pinging-google) to reindex the site regularly. That ping could be added to a crontask or elsewhere in a separate PR.

## Testing Instructions

 * Run the site locally following README instructions
* Visit http://localhost:8000/sitemap.xml and confirm you see XML with references to the two sitemaps described above
* Visit each and confirm that they contain the right content

### DIFF
--- a/committeeoversight/settings.py
+++ b/committeeoversight/settings.py
@@ -48,6 +48,7 @@ INSTALLED_APPS = [
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'django.contrib.sitemaps',
     'opencivicdata',
     'opencivicdata.core',
     'opencivicdata.legislative',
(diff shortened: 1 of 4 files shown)


beamalsky:
I was! Great news. I'll get this queryset ordered and then merge.

